PR #3, opened 2024-10-02
Added updated Website class to parse webpages which is designed using JavaScript 

Toulik-Das:
Added updated Website class to parse webpages which is designed using JavaScript heavily, at the end of week1 day1 notebook and tested the open AI summarization output using openai, cnn, edwarddonner webpages.

image
image
image


Toulik-Das:
Hi @ed-donner, let me know in case of any changes/addition need to be done.

PR #14, opened 2024-11-20
git ignore addd

manojpraba:


ed-donner:
Hey @manojpraba thanks so much for this.. can you give me some more background into why these are being added to gitignore - thanks so much

PR #39, opened 2024-12-15
Adding 2 projects to the community contributions.

shreshthkapai:
The 2 projects are as follows:
1) Research Paper summarizer using OpenAI API.
2) To-do list/Planner using Ollama model running locally.


ed-donner:
Wow! This is fabulous - thank you so much

PR #44, opened 2024-12-18
Correct typo in week4/day4.ipynb

cloudllama:
Change function from stream_code_quen to stream_code_qwen


ed-donner:
Wow how did I not catch that?! Thank you so much!

PR #46, opened 2024-12-18
Day 1 selenium challenged completed

codenigma1:
I added new class `WebsiteCrawler` and keep intact original class. Also, create similar function `new_summary` passing with additional parameter chrome_path which is require for selenium. Please check it. Thank you for teaching us!


ed-donner:
Brilliant!! Thanks so much for contributing this!

PR #48, opened 2024-12-19
Adding an example of implementing chatgpt.com in jupyter notebook

daimaukr:
**The lightweight and interactive ChatGPT interface for Jupyter Notebooks.** 
It enables users to send messages, view Markdown-rendered responses, and reset conversations.
Example of usage (model `gpt-4o-mini`):
image


ed-donner:
that's a really great idea - I love it - who needs to spend the $20 pro subscriptions when they have this!! Terrific

PR #49, opened 2024-12-19
Adding example that shows how to generate cover letter from cv

fletchertom:
Summarise a CV (resume.txt) then generate a cover letter from this CV - user will be prompted to enter a job description and this will also be used when generating the cover letter.


ed-donner:
Excellent project - thanks so much for contributing and sharing this!

PR #50, opened 2024-12-20
improve readability

gabormeresz:
A minor suggestion to make the code more readable


ed-donner:
Aha - well spotted - thanks so much for that

PR #51, opened 2024-12-20
Code generator using  gemini in windows setup

udayslathia16:


ed-donner:
Thanks so much - this looks great

PR #53, opened 2024-12-21
Week 1 Exercise

gopinath1998:
I have added my solution for the exercise.
I have little bit different solution for the ollama then learnt better solution from yours.


ed-donner:
Nice! Thank you so much!

PR #55, opened 2024-12-22
Two Nice Solution added; First, Select the best LLM, Second approach is collaborative.

codenigma1:
First solution I design to get response from your favorite LLM and second is colloborative approach of two LLM added might be good specimen for the upcoming student to think about it even better and good approach to refine this approach I leave it from them


ed-donner:
This is really high value - the multi-shot prompting is very clear, your treatment of the response is nice, and the collaborative approach is elegant and powerful. Really great stuff. Thank you so much.

PR #56, opened 2024-12-22
Added Windows capability to running cpp code

cloudllama:
- Add the ability to select the Visual Studio compiler for Windows using a radio button.
- Add the ability to populate the python text block with either the "pi" or "python_hard" code

### DIFF
--- a/week2/day4.ipynb
+++ b/week2/day4.ipynb
@@ -214,7 +214,7 @@
     "    response = {\n",
     "        \"role\": \"tool\",\n",
     "        \"content\": json.dumps({\"destination_city\": city,\"price\": price}),\n",
-    "        \"tool_call_id\": message.tool_calls[0].id\n",
+    "        \"tool_call_id\": tool_call.id\n",
     "    }\n",
     "    return response, city"
    ]


ed-donner:
Hey @cloudllama! This is fantastic!
I'm wondering about including the path to VISUAL_STUDIO_2022_TOOLS like that.. like do we know it will always be in:

C:\Program Files\Microsoft Visual Studio\2022\Community\Common Tools\VsDevCmd.bat

And is that the most common C++ compiler on a PC? Once of the drawbacks of having 20,000 students on the course is that there's a lot of different variations and people get frustrated when it doesn't work on their system. Some students don't use the C drive, some have older versions of software, etc.

How bulletproof is this? Can anything be done to make it more generic, or at least to add comments to the student about what to check for and what to install or change on their PC?

Alternatively: you could put this in a community-contributions folder, in which case students know to use it "as is". But if it's in the main week4 folder, then we need to check it will work in most conditions..

Thanks so much!

cloudllama:
@ed-donner Thanks for the fast response. Looking at the pull requests, I intuit that this is your full time job right now...I appreciate the sliver of attention. Of course, feel free to include or not; I simply offer any code as a service. (Insert here an ode to hands-on learning always being the most effective method to understanding.)

Visual Studio (Community Edition or Professional and not Visual Studio Code) are pretty much the most popular C/C++ compilers on Windows owing to their low, low price of free. I test for both the current version (2022) and previous version (2019). The path to the environment batch file (VISUAL_STUDIO...) can be considered static and immutable as Microsoft has no desire to move it around.

I've pulled the logic into a separate function that will detect the platform, test out the potential compilers, and return the compiler command that will work, or none if no compiler is found. Currently, the test includes:

Windows
- Visual Studio 2022
- Visual Studio 2019
- gcc

Linux
- gcc
- clang

- Mac
- gcc
- clang

I've run tests on Windows. I'm setting up an environment on Ubuntu to test the function under Linux. I don't have a Mac so I'll use your command line (for clang).

I expect I'll finish up tomorrow.

PR #57, opened 2024-12-22
Week 2 Day 4 - handle multiple tool calls

gabormeresz:
An alternate version of the week 2 day 4 code, which is able to handle multiple tool calls at once.
Example prompt to trigger 2 tool calls: "Could you tell me which plane ticket is cheaper, Berlin or Paris?"


ed-donner:
@gabormeresz terrific! Thank you so much

PR #58, opened 2024-12-23
LLMs 3 way solution including ollama

codenigma1:
Added solution where 3 llms are talking to each other with different perspective.


ed-donner:
@codenigma1 love it! Thank you once again!
"I think cats are better than dogs" 😂😂

PR #59, opened 2024-12-23
Removed Mac-specific binary

cloudllama:
Other platforms might accidentally try to run the Mac-specific binary, causing confusion.


ed-donner:
very good point @cloudllama  - I'll also add to the .gitignore

PR #60, opened 2024-12-23
Fix handle_tool_call function placement

Ishan-Karpe:
The function is now coming before the chat() function, as a function must be defined before calling/invoking it. This change prevents the "handle_tool_call" is not defined error.

### DIFF
--- a/week2/day4.ipynb
+++ b/week2/day4.ipynb
@@ -173,28 +173,6 @@
     "\n",
     "What we actually do is give the LLM the opportunity to inform us that it wants us to run the tool.\n",
     "\n",
-    "Here's how the new chat function looks:"
-   ]
-  },
-  {
-   "cell_type": "code",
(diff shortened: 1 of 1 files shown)


ed-donner:
Ah @Ishan-Karpe I very much appreciate your contribution.. but...   

I have made several blunders in my code in my time. But on this occasion, I don't believe I have an error! Unless I'm misunderstanding you, it's completely allowed (and very common) to call a function within another function before it's been defined. So this is valid python - try it out in a Jupyter notebook!

```
def function1():
    function2()

def function2():
    print("Tada!!")

function1()
```

And week2 day4 code runs nicely as is, as evidenced by the attached screenshot below - all cells have run, and the tool is called!  

If you were having problems, is it possible you didn't execute all of the cells?

Let me know if I've misunderstood.

order

PR #63, opened 2024-12-24
Call the appropriate C++ compiler on Windows, Linux, and Mac

cloudllama:
Here's the updated week4/day4. The final cell runs c_compiler_cmd function to determine the available C (C++, technically) compiler on the platform. The first compiler found, wins. If no compiler is found, then the button to run the C++ code is disabled.

The c_compiler_cmd attempts to compile a simple program. If the program compiles _and_ produces the expected output of "Hello", then we can be confident that the compiler is functional.

Tested on Win11 with Visual Studio 2019 & 2022. Tested on Ubuntu 24.10 with g++ and clang++. The format should be the same for Mac, although I don't have a Mac to test on.

If worthwhile, put into main. If just interesting, we can put it into community contributions.


ed-donner:
I am blown away @cloudllama - you've done a truly superb job with this. I'm going to need to make a major change to this though, in the form of adding a massive headline to call out your work here! Thank you so much - this is a really big improvement.

PR #64, opened 2024-12-24
Correctly identify Mac as a Mac not Linux

cloudllama:
Happy to help. I noticed that I identified a Macintosh as a Linux box. Copy-and-paste error and no Mac to test...


ed-donner:
Thank you @cloudllama - yes I saw that and was thinking it was intentional for some reason.. Gotcha.

PR #70, opened 2024-12-30
Update day5.ipynb

Ishan-Karpe:
4c can be unclear to some, 4 cents is better to use instead

### DIFF
--- a/week2/day5.ipynb
+++ b/week2/day5.ipynb
@@ -235,7 +235,7 @@
     "\n",
     "Let's put this in a function called artist.\n",
     "\n",
-    "### Price alert: each time I generate an image it costs about 4c - don't go crazy with images!"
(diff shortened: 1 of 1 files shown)


ed-donner:
@Ishan-Karpe thanks so much - good catch!

Ishan-Karpe:
@ed-donner does the PIL Libary exist? I can't download for some reason. it's not in requirements.txt, so it wasn't installed initially. Any alternatives to PIL?

ed-donner:
Hey @Ishan-Karpe oh that's odd, it should be installed as a prerequisite of other things in requirements.txt (you don't need to list all packages in requirements.txt, pip automatically installs prerequisities).
The package is called pillow.. 
pip install pillow
https://pypi.org/project/pillow/

PR #76, opened 2025-01-04
remove handle_tool_call hard-coded function call

mmaurier:


ed-donner:
thank you @mmaurier - great update

PR #79, opened 2025-01-06
PC Docs for using Mamba

kaviraj07:
### What:
- This PR provides documentation on using Mamba for creation of environment.yml on PC.

### Changes:
Changes made to Part 2 and a new part added, called Part 2C in the following:
- SETUP-PC.md
- SETUP-PC.pdf

### Why:
- Using Mamba is significantly faster than using Conda, which was stuck in resolving dependencies. Mamba, written using C++, provides faster installation time and dependency resolution.


ed-donner:
Oh wow @kaviraj07 this is super kind of you to contribute. I've been excited about the idea of adding mamba for some time.

One hesitation before I merge this in. It's amazing that I have 25,000 students on this course, but I've been a bit burned by how many different situations we run into. No matter how careful I am with the instructions, 1% of people will have an unexpected setup and will hit some roadblock. Have you had an opportunity to test the mamba generated environment on a PC and Mac by any chance? And with some of the labs? Otherwise I'll do it for sure, but it might take me some time.

Some more questions for you:
Once mamba has generated the conda environment, is it in other ways an ordinary conda environment? If you look at the troubleshooting notebook and the diagnostics scripts - do I need to modify them in any way to account for those who set up with mamba?

When mamba students do a conda env update, should they use a different command in place of conda env update -f environment.yml

I'm thinking of replacing Anaconda with Miniconda - will the mamba instructions work ok with miniconda too?

Should I consider just making mamba be part of 2A, i.e. just have it be the default that everyone uses?

Thanks again for your help on this - I would be so happy to get to a situation where everyone is able to reliably and quickly build a good environment, and mamba seems like it should be the answer!

kaviraj07:
Thank you @ed-donner for the feedback. I completely understand your concern about the variety of situations that exist. So I 'll try to answer to my best knowledge:

- I am currently using Windows as OS and therefore, can safely comment for the PC setup. So far, I do not have any issues, however, please note that I am still in the early stages of the course. I will be happy to provide further feedback again once I have progressed enough on the course.

- Yes Mamba behaves as a standard Conda environment and so far, I did not encounter any differences.

- Mamba students can still continue to use `conda env update` and will work totally fine. (If they need to use mamba for faster update, they can use `mamba env update -f environment.yml`)

- Yes Mamba works well with Miniconda and should not cause any difficulties. However, from the docs, it is mentioned to use the conda-forge channel as default channel.

- Making mamba as part 2A would be a good choice as it is faster and less resource intensive. This will provide a smoother experience for the students. But I think it is also helpful to retain the part of Anaconda (e.g. making it as Part 2B) as many of the students may be more familiar with that setup too.

So in my opinion, this pull request can remain open for now, untill I have tested the other weeks of the course. I will be able to comment again on the PC side but unfortunately not for Mac.

Thank you again for this consideration.

ed-donner:
That's super clear @kaviraj07 thank you so much. I'll do some testing too in the coming weeks. Many students are finding that Anaconda gets stuck, and I'd be so happy to move to something faster and simpler...
Keep me posted please!

PR #81, opened 2025-01-06
chore: add missing dependencies for non-Anaconda users

sorenandersen:
Couple of deps were only present in `environment.yml`, this tiny change adds them to `requirements.txt` in order to use standard Python virtual env's.


ed-donner:
Oh gosh - thanks so much @sorenandersen - thank you for spotting these and taking the time to add to requirements.txt

PR #85, opened 2025-01-07
My Attempt at Code Docstringer and Commenter

oidebrett:
A code tool that automatically adds docstring / comments. Keeping it simple. Used LLM to generate docstrings and code comments. I checked with an LLM to give me the most effective system and user prompt.


ed-donner:
Brilliant @oidebrett thank you so much! Terrific project

PR #108, opened 2025-01-18
Week 4: Gen tool for writing unit test cases

elena-shirokova:
Screenshot 2025-01-18 at 12 09 58

## Current flow:

1. For a python code it generates the unit tests using `pytest` library. The dashboard supports tests execution along with a coverage report. If the unit tests are fine, there is an option to save them for future use. It can happen, especially with Ollama , the tests are having a typing error. In this case the code can be edited in the right window and executed afterwards. 

2. Supports 3 models: 

- gpt-4o-mini
- claude-3-5-sonnet-20240620
- llama3.2

It is recommended though to use other models except Ollama, my tests showed the code returned from ollama required more supervision and editing. Some generated unit tests from ollama don't provide full coverage, but still it is a good starting point for building such a tool..


ed-donner:
Beautiful project @elena-shirokova thank you so much. A valuable example.

PR #112, opened 2025-01-20
barry_northern: using Gemini SDK instead of Claude + use Gemini's chat function.

barrynorthern-tf:
This is a reworking of the initial chat exercise on week 2 day 1. The original uses GPT and Claude, whereas this one eschews Claude for Gemini.

I found that Claude has to be set up with billing before you can call the API, whereas Gemini does not.

Gemini also has a [nice chat function in its Python SDK](https://ai.google.dev/gemini-api/docs/text-generation?lang=python#chat) which we use here.

### DIFF
--- a/week2/community-contributions/day1-conversation-with-gemini.ipynb
+++ b/week2/community-contributions/day1-conversation-with-gemini.ipynb
@@ -1,0 +1,616 @@
+{
+ "cells": [
+  {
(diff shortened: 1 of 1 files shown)


ed-donner:
Brilliant @barrynorthern-tf thank you so much for the contribution

PR #115, opened 2025-01-22
Added my day1 contribution to community-contributions

md-Junaid:
1. This version will Analyze your CV against the job description and if the CV is compatible with the job then it will write a **cover letter**.  It will also give a short summary of how the CV is fit or not fit for the job.

2. The job description is fetched using a URL, so you need to paste the URL instead of the job description text.

3. If either the job URL or CV is just random text or fake, then it will not write a cover letter instead tells you to enter the correct information.


ed-donner:
brilliant, @md-Junaid - thanks so much for this. A great business area for LLMs (as I know well!)

PR #117, opened 2025-01-24
barry_northern: booking function and multiple tools calls

barrynorthern-tf:
- the multiple tool fix is slightly different to another contribution, using a functional approach.
- without the fix you get an error about the fact that all tool ids must follow in messages that request them.
- this occurs when, for example, you ask for prices for two cities.
- the booking function and prompts allow you to book at a different price if you ask explicitly.
- this can be fun if you start asking about the discount applied.

Screenshot 2025-01-24 at 22 32 26


ed-donner:
very nice - that's an elegant solution - thanks so much @barrynorthern-tf

PR #123, opened 2025-01-27
Added day 2 exercise to community contributions 

neilneyman:
including a tweak to disable SSL verification in the BeautifulSoup website call.

I also commented about this in the Q&A section of the course. I included this in my answer to be able to run the course materials behind a zscaler VPN that changes the SSL certificates.  This may be a somewhat common issue for zscaler users.

SSL verification should only be disabled for classroom/testing purposes, this is not recommended for production code.


ed-donner:
Thanks so much @neilneyman - I've added some notes on this to the troubleshooting guide too - super helpful, and great example

PR #126, opened 2025-01-28
Feature: Added code and modules to create doc string for any python c…

narquette:
# Overview

Created a module that will read in an existing python file and create a new file with a LLM model suffix that will include doc strings for all functions listed in a file.


ed-donner:
Wow @narquette that's like a legit valuable commercial tool right there. I might use this myself in my day job! I hope you do too!

narquette:
I appreciate the kind comments.

Thanks,

Nick Arquette

PR #129, opened 2025-01-29
Update SETUP-linux.md

MaxGandini:
This is a little extra for the pip method using distributions such as arch. This should come in handy for someone that is using rolling release distributions as generally dependencies are broken and troubleshooting takes a long time when done alone.


ed-donner:
oh wow thank you so much @MaxGandini this is really valuable info & makes the linux setup more resilient. V much appreciated. Gensim isn't an important part of the class, no problem to remove it

PR #132, opened 2025-01-30
Week 2 day one exercise for community.

neilneyman:
Added my version of a 3-model conversation, the models will discuss oscar nominations from optimistic, pessimistic, and mediator perspectives.


neilneyman:
Also added Week 2 Day 5 exercise.   There are 2  versions submitted in this exercise.  

The first  uses 3 tools, one is to show ticket prices, another for departure times by city, and a third tool to complete booking, that takes the  user's name, destination city, and departure time.

The second chat example adds the translation agent, with a drop-down to select the language.

ed-donner:
Fantastic work @neilneyman thank you so much for this valuable contribution!

PR #139, opened 2025-02-01
Added test case automation solution

samt07:
Hi Ed, I did a small project to automate test cases from requirements doc. I also did the Week 1 exercise, but didnt want to upload as i see many have done it already :)


ed-donner:
Terrific, @samt07 - this is a very nice addition. Thank you!

PR #145, opened 2025-02-02
added installation of code auto completion and git for jupyter 

amirkolaman1:
added installation of code auto completion and git for jupyter  inside requirements.txt and environment.yml

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -48,3 +48,6 @@ dependencies:
     - feedparser
     - twilio
     - pydub
+    - jupyterlab-lsp
+    - jedi-language-server
+    - jupyterlab-git
--- a/requirements.txt
+++ b/requirements.txt
@@ -42,3 +42,6 @@ setuptools
 speedtest-cli
 sentence_transformers
 feedparser
+jupyterlab-lsp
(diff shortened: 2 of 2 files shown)


ed-donner:
Hey @amirkolaman1 - these are both really valuable additions.

I might need to take a while before I merge this, though. It's a great joy that I have 40,000 students on the course! But the downside is - nothing prepared me for how unpredictable and fragile different environments are. Any time I make an infrastructure level change, even if it seems harmless, about 1 in 100 people hit an issue. Whether the 260 character limit in Windows file paths surfaces, or local admin permissions, or anti-virus software flares up - something weird happens.

And that means I get 400 messages from frustrated people!

Someone else contributed a great change of using Mamba instead of Anaconda, which I'm interested in, but again I'm nervous about what might result..

So I'll need to test these extensions locally and get comfortable with them for a while. I notice that the jupyterlab git extensions has some troubleshooting stuff on pypi. If you've got experience with them and have any thoughts on whether this is pretty reliable for me to add, I'd v much appreciate it!

Thanks so much for adding these.. they will be very useful
Ed

PR #152, opened 2025-02-06
Feature/add deepseek and new hf inference provider option to w4d4cc

atwmarshall:
Adds functionality to use new HuggingFace capability - Inference Providers - which allows free access of certain models without need to spin up paid for HF Inference Endpoints.  

This means you can run Deepseek easily.


ed-donner:
That's super helpful @atwmarshall thank you for this! Very valuable.

PR #156, opened 2025-02-08

Lukid:
Added website summarizer using local Llama 3.2 (via Ollama) with Italian prompts

This pull request integrates the code from day1 to build a website summarizer that leverages a locally running Llama 3.2 instance instead of OpenAI’s model. The implementation uses Italian prompts and employs the OpenAI API to call Llama 3.2 locally.


ed-donner:
Great work - thank you @Lukid !

PR #158, opened 2025-02-08
Fixed an issue where compiler_cmd was undefined.

motoya0118:
Hi, Ed!
I'm Motoya, an engineer from Japan.
I'm really excited every day taking your course!
Thank you so much as always!

By the way, while going through the course, I ran into an error because a list declaration was missing.
I’ll be sending a pull request with the fix, so please take a look!

### DIFF
--- a/week4/day4.ipynb
+++ b/week4/day4.ipynb
@@ -480,8 +480,9 @@
    "source": [
     "def execute_cpp(code):\n",
     "    write_output(code)\n",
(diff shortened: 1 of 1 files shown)


ed-donner:
Thanks so very much @motoya0118 - well that's so curious, it does look like an error in there. I can't imagine how that happened or how it ever worked. Thank you so much for fixing this - hugely appreciated.
Ed

PR #163, opened 2025-02-09
Week2-Day5-Airline flight booking

wjm8823:
Project - Airline AI Assistant
I've added database connectivity to enable Openai to:

Retrieve ticket prices
Display the number of available seats for each flight
List all available destination cities
Facilitate seat bookings
Once a booking is confirmed, an image of the booked destination city is displayed.


ed-donner:
Thanks so much @wjm8823 - great example of adding tools functionality

PR #164, opened 2025-02-09
Jupyter notebook to ask a question and get a response from the GPT, Claude, and Gemini

miptirdis:
In my example, a user can ask a question and get a response from the GPT, Claude, and Gemini. You don't need to select a model and re-type the question. It might be helpful to see all responses on one screen and pick the best one :-)


ed-donner:
Simple and effective! Great work, @miptirdis.

PR #166, opened 2025-02-10
Gradio UI with recording voice and translating it

petria-tieto:
translating audio recorded to a file


ed-donner:
A valuable project @petria-tieto thank you so much for the contribution.

PR #173, opened 2025-02-14
Week1 Day 1 - Implement simple election program q&a tool (DE)

Phi-Li-Ne:


ed-donner:
Thanks for adding this example @Phi-Li-Ne

PR #175, opened 2025-02-14
added day4 exercise with inference provider

FireEagle97:
Added the use of inference providers that was introduced recently by Hugging Face to convert the code. Improved the user prompt to include algorithmic efficiency and performance optimization.


ed-donner:
This is super useful - thanks so much @FireEagle97 - and great investigation on the prompt to fix the algorithm

PR #185, opened 2025-02-17
Adding Ollama brochure version and Ollama Tutor version with streaming

johnIT56:


ed-donner:
Thanks so much @johnIT56 - very valuable contribution

PR #186, opened 2025-02-17
Week1 Exercise Ollama Solution with Streaming

HamzaSH6:
**Used Ollama and Added Streaming for Llama Models.**


ed-donner:
Nice! Thank you again @HamzaSH6

PR #191, opened 2025-02-18
Useful to have conda env. available in jupyter lab

rcuevass:
Upon creating the conda virt. env, llms, I think it useful to have that environment available within the list of kernels once jupyeter lab is launched. I added few lines in SETUP-PC.md describing the process.

### DIFF
--- a/SETUP-PC.md
+++ b/SETUP-PC.md
@@ -76,6 +76,13 @@ If this Part 2 gives you any problems, there is an alternative Part 2B below tha
 
 You should see `(llms)` in your prompt, which indicates you've activated your new environment.
 
+**Important**. Before launching Jupyter Lab (section below), it is useful to have the `llms` conda environment just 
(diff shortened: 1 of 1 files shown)


ed-donner:
hey @rcuevass thank you for this. Help me understand this - for me, if I activate the llms environment with:
`conda activate llms`
and then in the Anaconda Prompt, I launch jupyter lab with:
`jupyter lab`
Then the kernel runs in the activated environment, without needing to install anything else or select the kernel. Does that not happen for you?
Thanks so much

rcuevass:
hey @ed-donner . If you launch jupyer from the the conda env `llms` one would think the kernel runs in the activated env. I remember that not wokring for me in the past in other projects (specially on Windows), thus I have made what I proposed in the README a practice. 

In addition, with this approach one could launch jupyter in the `base` env and then select the `llms` kernel at the jupyter lab level.

Hope that helps

PR #192, opened 2025-02-19
Add week2 with openrouter.ai and day 2 brochure practice

stoneskin:
with openrouter.ai,  can access all LLM with openAi sdk,
And there are many free mode you can use include google, meta and deepseek, etc


ed-donner:
Hey @stoneskin this is really tremendous - so useful for everyone to have an openrouter example. Massive thanks.

PR #206, opened 2025-02-22
Made a 3-way chat for GPT, Claude, and Gemini

fireonehero:
I added a 3 way chat system between GPT, Claude, and Gemini!


ed-donner:
Hey @fireonehero - actually your PR appears to change the original lab. Would you be OK to update the PR so it leaves the original lab as is, but adds your version to community-contributions? Thanks so much
Ed

PR #208, opened 2025-02-22
Adding Ollama with tts and use it as translator

johnIT56:


ed-donner:
Brilliant, @johnIT56 - thank you so much!

PR #209, opened 2025-02-22
Mac lama selenium

hassanbsee2071:
Faced some challenges while setting up selenium for the mac.


ed-donner:
@hassanbsee2071 thanks so much for sharing - excellent

PR #222, opened 2025-02-26
gemini-codes-week5

serpentile-c137:
I have tried doing day 3 and day 4 of week5 Rag pipeline using Google Gemini


ed-donner:
thanks so much @serpentile-c137 - really valuable to get Gemini versions in here. Thank you for taking the time to share with others.

PR #226, opened 2025-02-28
Completed Exercise day 2

mapatro:
Completed first exercise


ed-donner:
hey @mapatro it looks like your PR is changing an existing file in community-contributions. Was that intentional? If not, you might want to rename your contribution so it doesn't overwrite the existing one? Thanks so much
Ed

mapatro:
Should I rename to different nam3?

Mahesh

ed-donner:
Yes please @mapatro - if you could rename it that would be much appreciated. Thanks so much

PR #228, opened 2025-02-28
contribute to week 2 examples with a tool: python interpreter

flacout:
Added an example in Week2 community-contributions:
python interpreter as a tool, with multi shot prompting.


ed-donner:
Wow this is a pro example @fabrice-lacout. Quite something to give your LLM the ability to exec() ! Very interesting indeed. Thanks for making this tool and for sharing.. fascinating. And, a bit scary!
Ed

PR #238, opened 2025-03-04
add ollama with streams using ollama generate

banuady:
As part of Week 1 exercise, for the ollama calling function I used `ollama.generate()` instead of `ollama.chat()`. It also uses streams, although it was not required for this challenge.

This was based on the examples on the ollama-python repository: https://github.com/ollama/ollama-python/tree/main/examples.


ed-donner:
Thank you so much @banuady - very nice example, and always good to use streaming!

PR #247, opened 2025-03-09
Added frontier model testing table compared to basic models and human…

kellewic:
I tested numerous frontier models from OpenAI, Anthropic, Google, and others via Groq API and this notebook shows the results.


ed-donner:
I'm so happy you added this @kellewic - these results are great - thank you again

PR #251, opened 2025-03-11
Request review of the day1_with_selenium-CB.ipynb

Chiranjib1979:
. It seems to be broken but I am not able to fix the issue. Will be great help if you could resolve


ed-donner:
Hey @Chiranjib1979 - thanks for sharing.
What's the problem with it, and have you made any progress with debugging?
Perhaps you could send me an email to ed@edwarddonner.com with a bit more on the issue and your investigation so far. Have you looked at the other Selenium implementations in community-contributions?
Thanks
Ed

PR #252, opened 2025-03-11
Adding week2 exercise work with a translator using different frontier…

manish-gajria:
… model and also trying audio input using whisper library

Wrote the translator function to take in the full chat history, breaking it down into individual messages, sending them into model as a "user" role message, recieving the translation as a response, and then reconstructing the message with the original role so gradio can display the translated message in the correct order. Previous attempts results in a lumped translation of the entire history showing up as an "assistant" response in gradio.

### DIFF
--- a/week2/community-contributions/week2_day4_exercise.ipynb
+++ b/week2/community-contributions/week2_day4_exercise.ipynb
@@ -1,0 +1,408 @@
+{
+ "cells": [
+  {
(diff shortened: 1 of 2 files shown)


ed-donner:
That's super interesting and a valuable example @manish-gajria - nicely experimented and built. Thank you for sharing this.

PR #258, opened 2025-03-13
Correction in week2/day1.ipynb

MartijnR:
correction in the deepseek response length check


ed-donner:
Ah silly me. Thank you so much @MartijnR - nice catch. Thanks for fixing & contributing.

PR #261, opened 2025-03-15
Changed to use modal.Volume instead of modal.build()

kellewic:
Removed modal.build() in favor of using modal.Volume to cache model files.
Added warming code as an app.function().

### DIFF
--- a/week8/pricer_service2.py
+++ b/week8/pricer_service2.py
@@ -3,17 +3,20 @@ from pathlib import PurePosixPath
 
 # Setup - define our infrastructure with code!
 app = modal.App("pricer-service")
+secrets = [modal.Secret.from_name("huggingface-secret")]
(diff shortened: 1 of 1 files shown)


ed-donner:
@kellewic this is so brilliant - I'm super grateful. Would it be possible to remove the pricer_wake_up change? In the spirit of changing as little as possible (anything I change always seems to break 100 people!) and also being cautious not to accidentally cost a student more than they're expecting. If you're able to make that change to the PR, I'd be super grateful. This is really excellent.

kellewic:
@ed-donner I removed the wake code and added min_containers with a default of zero. This won't keep any containers alive by default. Changing to 1 will keep a single container alive. I need to take a class on git/github - I had to figure out how to change a PR 😃

PR #269, opened 2025-03-20
Llm week5 community: RAG with Ollama used to query documents

petria-tieto:
I must say that this does not work as expected. I have submitted this PR though, maybe you @ed-donner  have a helpful hint?
I have about 600 files in one folder and want to query them. 
I thought this to be a trivial task, but it was not 😃
Even though the retriever has the information required, it's dropped from the answer. How should this be done so it finds the document and adds the

Here is an example where LLM stumbles on two documents and gives misleading answer.

query = "How to fit the cassette joint to a Shimano Nexus 7 Speed Hub?"
(there is a document exactly with that title)

from a callback:
Screenshot 2025-03-20 at 13 53 36

_Answer: The answer to the question "How can I fit the cassette joint onto a Shimano Nexus 7-speed hub?" is not provided in the text snippet. The text appears to be a tutorial or guide on how to convert a recumbent bike to an electric bike, but it does not provide instructions on how to fit the cassette joint onto a Shimano Nexus 7-speed hub.

However, based on general knowledge of bicycle mechanics and repair, here is a possible answer:

To fit the cassette joint onto a Shimano Nexus 7-speed hub, you will need to follow these steps: ..._

Further difficulty for the AI: not all documents are in English?

### DIFF
--- a/week2/community-contributions/day4_compare_prices.ipynb
+++ b/week2/community-contributions/day4_compare_prices.ipynb
@@ -1,0 +1,275 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "id": "ddfa9ae6-69fe-444a-b994-8c4c5970a7ec",
+   "metadata": {},
+   "source": [
+    "# Project - Airline AI Assistant\n",
+    "\n",
+    "We'll now bring together what we've learned to make an AI Customer Support assistant for an Airline"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "8b50bbe2-c0b1-49c3-9a5c-1ba7efa2bcb4",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# imports\n",
+    "\n",
+    "import os\n",
+    "import json\n",
+    "from dotenv import load_dotenv\n",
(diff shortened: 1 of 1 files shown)


ed-donner:
Ah yes @petria-tieto that looks like an easy one. You're using llama3.2 right? The bottom line I believe is that you're out of the performance capabilities of a tiny model like llama3.2. At 2-3 billion parameters, it is several thousand times smaller than the mighty GPT-4o - many orders of magnitude. I think it's a similar size to GPT-2! The fact that it's able to be coherent with simple english is something of a miracle; but stretching it with comprehension like this is just too much. You can demonstrate that by temporarily using a frontier model, and you should find it has an easy time.

You can reproduce this yourself by doing "fake RAG" - simply construct a prompt that looks a bit like that full text above and includes the context and the question. Then do on the command line:
`ollama run lllama3.2`
And try out the prompt. I suspect it will struggle.
Then go to ChatGPT, and I imagine it will have an easy time.
You could try some medium sized models to see how they handle it.
Let me know!

By the way, you can submit the PR for real if you'd like - just include a note at the top that you're still having problems getting a good response, and other people can try experimenting too.

Thanks for the R&D - brilliant!
Ed

PR #271, opened 2025-03-21
The Airline AI Assistant can now compare prices

simonsteinberg:
The Airline AI Assistant can now compare prices, e.g., 'Is it cheaper to go to Berlin or to go to London?'


ed-donner:
That's a great addition - brilliant, @simonsteinberg, thank you!

PR #274, opened 2025-03-24
Update requirements.txt

dnyanesh-github:
Add "notebook" to the requirement.txt file to enable the command "jupyter notebook"

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -42,3 +42,4 @@ setuptools
 speedtest-cli
 sentence_transformers
 feedparser
+notebook


ed-donner:
hey @dnyanesh-github - hmmm what's your rationale for this? I don't think it's needed for the course? I'm reluctant to add anything to requirements.txt, because any dependency I introduce (however harmless it should be) seems to negatively affect about 1% of people. And with 60,000 on the course, that means I get 600 angry emails!

PR #278, opened 2025-03-25
Create day5_openai_whisper_llamainstruct

palbha:
Added code to use Opensource - Openai whisper tiny model & deploy a gradio app which take audio input or audio file


ed-donner:
That's an excellent addition, thank you so much @palbha

PR #279, opened 2025-03-25
Code Generator using OpenAI and Ollama

yashupatel1998:
I have completed Day1 and Day2 exercise and implemented code generator which identifies responce difference between Openai and Ollama.


ed-donner:
Thank you so much for this @yashupatel1998 - it looks like you may have checked in some outputs from other labs - would it be possible to clear these outputs and push those changes to your PR? Thanks so much

PR #284, opened 2025-03-27
introduces groq cloud

sameer2407:


ed-donner:
Hey @sameer2407 - would you be able to move the changes into community-contributions for me? I prefer not to change the main labs as this can cause changes while students are progressing..

PR #285, opened 2025-03-27
Update to environment.yml for Pytorch

steveroberts1980:
Pytorch no longer has support for conda packages. Moved pytorch to pip.

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -10,7 +10,6 @@ dependencies:
   - numpy
   - pandas
   - scipy
-  - pytorch
   - jupyterlab
   - ipywidgets
   - matplotlib
@@ -48,3 +47,6 @@ dependencies:
     - feedparser
     - twilio
     - pydub
+    - torch
+    - torchvision
+    - torchaudio


ed-donner:
Hey @steveroberts1980 

Hmm are you sure about that? My understanding is that pytorch moved to conda-forge, and you'll see in the environment.yml that conda-forge is the first channel specified.

I just deleted and recreated my conda environment from scratch, and it created fine, including PyTorch.

Sometimes pytorch gives a problem installing on PCs that do not have the Windows 260 character limit constraint removed - see the "gotchas" section at the top of SETUP-PC.md.

I'm reluctant to make a change to something this fundamental in the environment setup.. if conda isn't installing pytorch, then this could affect the versions that it picks for all other dependencies. If I make a fundamental change to packages, this often negatively impacts between 1% and 10% of the students on the course, who experience anomalies with their setup. And with 60,000 people on the course, that means I can get hundreds - or even thousands! - of frustrated emails.

So I'm quite cautious about changes like this!
Let me know what you think.
Ed

PR #296, opened 2025-04-03
Day1ChandanUpdates

AvgAIJoe:
Checked and ran the code with different website.Also, checked the messages for openAI using system and user prompts.


AvgAIJoe:
Tested and checked

PR #301, opened 2025-04-07
Day1: Grafana Dashboard Metrics Summarizer

thisforcode:
Simple Jupyter Notebook that simulates a Grafana-like dashboard using mock system metrics (CPU, memory, GPU, disk errors), and uses a GPT model to summarize the data. It does not use Grafana itself, but mimics the concept for GPT integration.

**Example Output:**
- Average CPU usage is 65%, within normal range.
- Memory usage is at 13.2 GB of 16 GB, approaching threshold.
- GPU usage is at 72%, no bottlenecks detected.
- Disk read errors recorded: 3 (no action needed).

System is stable. Monitor memory for potential increase.

### DIFF
--- a/week1/community-contributions/mock-dashboard-summary/day1-dashboard-metrics-summary.ipynb
+++ b/week1/community-contributions/mock-dashboard-summary/day1-dashboard-metrics-summary.ipynb
@@ -1,0 +1,183 @@
+{
+ "cells": [
+  {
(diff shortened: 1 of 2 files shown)


ed-donner:
That's really excellent @thisforcode - great example of synthetic data and a nice use of LLMs. Thanks so much for sharing this.

PR #306, opened 2025-04-08
Add eCommerce Assistant notebook for product price retrieval week2

ariel1985:
I used a lot of "copy past" from your code and adapted with with chatGPT but it works!! 

Something I've noticed, I'm either missing something or llama3.2 is really bad at running tools. 
I've tried multiple times, different conversation scripts but even writing a simple "Hi"  get's the model confused

I added a backdoor in the system prompt to tell me what model it's running on


ed-donner:
Thanks so much @ariel1985 - yes llama3.2 does struggle with tools, and I find that it always calls them. I've heard that one solution is to provide a tool for the final answer. And keep tools as simple as possible..
Great experiments!

PR #308, opened 2025-04-10
Added Perl to Python Code Converter

samt07:
Hi Ed, I tried doing the code converter for Perl script to Python. Added here.


ed-donner:
Terrific @samt07 - that's an excellent addition with real practical application. Thanks for sharing..

PR #331, opened 2025-04-18
Voting Bots - multi-modal chat app

msource1:
A multi-modal (text, image, audio) chat app that leverages multiple tools-calling + several Gradio features, including the State component.


ed-donner:
Fantastic @msource1 - thank you for contributing!

PR #336, opened 2025-04-21
added youtube transcript summarizer

kbjerring:
Added YouTubeWebLink class instead of Website with simple youtube video transcript summarization capability


ed-donner:
Thank you @kbjerring - a nice and simple change with an impressive result! Great stuff

PR #341, opened 2025-04-23
Week 1, Day 1: pluggable architecture for web scraping and summarization

khalid-taha:
This system implements a **pluggable architecture** for web scraping and summarization, built on interface-based design using Python’s `Protocol` types. Each stage of the pipeline—content fetching, HTML parsing, and LLM-based summarization—is defined through explicit structural contracts rather than concrete implementations. Components like `RequestsFetcher`, `RobustSoupParser`, and `OllamaClient` fulfill these protocols and can be swapped independently, enabling flexibility, testing, and future extension without modifying core logic. Immutable data models (`@dataclass(frozen=True)`) enforce data integrity throughout the pipeline, while the design cleanly separates concerns across modules to support maintainability and modular growth.

### DIFF
--- a/week1/community-contributions/day1-pluggable_scraper_summary.ipynb
+++ b/week1/community-contributions/day1-pluggable_scraper_summary.ipynb
@@ -1,0 +1,431 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "id": "35f59eb3",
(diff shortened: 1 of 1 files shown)


ed-donner:
Thank you @khalid-taha - a very elegant design, and a useful example for others

PR #353, opened 2025-04-26
Community contributions branch

massalinux:
Added an example about Footbal Game(soccer for us people)


ed-donner:
hahaha brilliant @massalinux - football for sure! Thanks so much for a fun & valuable addition

PR #355, opened 2025-04-26
having some fun with "fitness". 

ckujawa:
As a fitness enthusiast and CPT (in addition to being a software engineer) I decided to have a bit of fun based on an old SNL skit. The response I just got surprised me...and was quite funny.

:muscle:


ed-donner:
Hahaha that's inspired @ckujawa - thank you for making this and sharing

PR #356, opened 2025-04-28
Added week2 day1 exercise - fictional chat between 3 actors

dan-palade:


ed-donner:
Terrific - thank you @dan-palade for this

PR #367, opened 2025-05-07
my community contribution for week 2 day 1

evokateur:
Ishmael, Starbuck, and Flask (no relation) on the Pequod? This is useful!


evokateur:
After going round and round with ChatGPT it finally coughed up the fact that you can create a branch from `upstream/main` to do these things on (which is what I did)

ed-donner:
Hahaha I had a temporary blank and had to google to remind myself what this was about... inspired! Thank you @evokateur . Oh yeah.. is that not explained properly in the PR guide? I can make another if it should be clearer

PR #368, opened 2025-05-07
Added tool call to course material seach

kbjerring:
This one proved a bit more complex than I had thought. A bit of a messy solution. Would certainly appreciate some tips on how to better organize streaming with tool calling in a clean way :)


ed-donner:
oh wow that's a cool idea @kbjerring - I'm afraid streaming with tool calling seems to be a super messy thing, unless you separate out the tool steps as 2 distinct LLM calls. If you ever look at OpenAI Agents SDK, they've wrapped this into something rather easier..
Thanks so much for this - what a great project

PR #371, opened 2025-05-09
Community contributions branch for week 1

markvenn666:
An email checker that makes emails more professional, does a spell check into UK English and creates a suitable email subject from the content


ed-donner:
Thanks @markvenn - great!

PR #375, opened 2025-05-10
A LLM summarizer of websites built with Pyppeteer in order to have mo…

Execotryx:
A LLM summarizer of websites built with Pyppeteer in order to have more streamlined run, than Selenium. Pyppeteer will download Chromium on its own - no need to have Chrome installed. Handy when you don't have UI.


ed-donner:
Terrific - thank you @Execotryx !!

PR #386, opened 2025-05-15
fix dumb mistake in (week 2) community contribution, fix Ishmael (he is not Ishmail, but it is hard not to type "mail"), remove debug cells

evokateur:
The dumb mistake was printing the GPT (Flask) message as the Gemini (Ishmael) message. Now it makes as much sense as it possibly can, I think.


evokateur:
Also, both GPT and Gemini take the hint and prefix their messages with their assumed name but Claude does not, only occasionally including `** as Starbuck **`

evokateur:
Ishmael, what a know it all! https://ishmael.app/

ed-donner:
hahaha thank you @evokateur !!

PR #389, opened 2025-05-16
Company Brochure Creation using Ollama

sharathir:


ed-donner:
hmm this seems to have no changes associated with it @sharathir - is that because I merged another PR from you? Should we close this one?

sharathir:
> hmm this seems to have no changes associated with it @sharathir - is that because I merged another PR from you? Should we close this one?

Yes please, I am also not sure how it happened! Thank You!

PR #400, opened 2025-05-23
add alternate week 4, day 4 which works on windows

chaser1802:
This will run week 4, day 4 end-to-end on windows OS. User only needs visual studio build tools.


ed-donner:
Thanks so much @chaser1802 - much appreciated

PR #405, opened 2025-05-25
job recommendation based on the resume content.

Shikidwaja:
- Job recommendation based on the resume content. Suggests the job role to apply for and provide the links from job sites based on location and sites user wants to search for.


ed-donner:
Great stuff @Shikidwaja ! Thanks for contributing. This is a business area that I know a thing or 2 about...

PR #443, opened 2025-06-08
Delay initial yield to avoid raw Markdown rendering in Gradio

axmac:
Gradio's Markdown output can render as raw text if the first few streamed tokens form incomplete syntax (e.g. unclosed bold or headers). This change adds a buffer before yielding to ensure the first chunk is syntactically valid Markdown.

### DIFF
--- a/week2/day2.ipynb
+++ b/week2/day2.ipynb
@@ -318,8 +318,12 @@
     "    )\n",
     "    result = \"\"\n",
     "    for chunk in stream:\n",
-    "        result += chunk.choices[0].delta.content or \"\"\n",
-    "        yield result"
+    "        next_token = chunk.choices[0].delta.content or \"\"\n",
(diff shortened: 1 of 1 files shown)


ed-donner:
hmm @axmac could this not result in the last chunk not getting yielded?

And if the text after 50 characters is in the middle of a markdown tag, wouldn't you get the same problem?

I guess I'd make the case that the existing code would only flicker for a moment before it corrects itself, so the simplicity of yielding as we go seems to outweigh the benefit of trying to detect markdown tags?

axmac:
I was getting very consistent failures last weekend, and now I can't reproduce it. I'll put it down to a glitch, I agree the simple approach is best, especially for a learning exercise. It was an interesting diversion to find a solve for the problem that I was seeing. :)

PR #444, opened 2025-06-08
Add dev container for streamlined development and switch to Mamba for faster environment setup

bvsatyaram:
This PR introduces two key enhancements to improve the learner experience:

1. **Devcontainer Support**
    - Adds a .devcontainer directory with the necessary configuration to enable development using [Visual Studio Code Dev Containers](https://code.visualstudio.com/docs/devcontainers/containers).
    - Provides a consistent and reproducible development environment with pre-installed dependencies.

2. **Switch to Mamba**
    - Replaces conda with [mamba](https://github.com/mamba-org/mamba) for faster and more reliable environment resolution and installation.
    - Speeds up the setup process, especially beneficial in CI and container-based workflows.
    - This effects only the devcontainer setup

These changes aim to reduce setup time and make it easier for contributors to get started with development on the project.

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -30,8 +30,12 @@
 
   // Use 'postCreateCommand' to run commands after the container is created.
   // Create .env file with placeholder for OPENAI_API_KEY
-  "postCreateCommand": "[ ! -f .env ] && cp ./.devcontainer/.env.example .env && echo 'Please set your API keys in the .env file.' || true"
+  "postCreateCommand": "[ ! -f .env ] && cp ./.devcontainer/.env.example .env && echo 'Please set your API keys in the .env file.' || true",
 
   // Comment out to connect as root instead. More info: https://aka.ms/vscode-remote/containers/non-root.
   // "remoteUser": "vscode"
+
+  // Use 'forwardPorts' to make a list of ports inside the container available locally.
+  // Forward a range of Gradio ports (7860-7879).
+  "forwardPorts": [7860, 7861, 7862, 7863, 7864, 7865, 7866, 7867, 7868, 7869, 7870, 7871, 7872, 7873, 7874, 7875, 7876, 7877, 7878, 7879]
 }


ed-donner:
Hey @bvsatyaram 

I'm really grateful that you did this work and shared it, and from a review it looks terrific. I'll keep looking at this. But I have to be honest: it's unlikely that I'll merge this in.

I need to brief you on some important context:

There are almost 100,000 people taking the course. You would not believe the number of different types of computers and OSes that people use. From old Windows installs, Intel Macs with old MacOS, different Linux distros, even some people working on Tablets. It's wild.

As a result, I get an enormous number of environment related questions and problems. Out of the 300 questions a day that I get, perhaps 100 of them are environment related. The Windows 260 character limitation. No MS Build Tools or Apple Developer tools. People mistyping "environment" and "OPENAI". ISPs throttling downloads. Old MacOS not supporting modern packages. Mismatches in C++ compilers.

So it's been important to focus on consistent, simple instructions for everyone, that are battle-tested on all systems to be as stable as possible, and that I can support. If I add Devcontainer support, then I guarantee you there will be a new bunch of questions (yes, even though it should be reproducible) - perhaps 20 a day - and I won't have tested it in all situations and I won't be able to guarantee it works.

And for sure Mamba is a nice way to do it; other students have used Mamba; actually other students have raised PRs for Mamba, some pyenv, some poetry - they all have their pros & cons. One student wanted to contribute a Docker container setup. If I merged all these PRs, there would be another slew of ways to create your environment.. and all the associated issues.

And it's not just about the environment setup! We'd need to check that all the labs run smoothly in the new environment, for all weeks, in Windows, Mac and Linux. There's a lot to test.. When I first made the Anaconda environment, I tested it extensively for all labs in all major platforms.

I'm not a massive fan of Anaconda, but it has the benefit of being the OG of data science environments, and when it works, it works. When it doesn't, I prefer the 2B alternative of just using virtualenv, because it's the simplest fallback.

I am interested in updating the repo to use the fantastic uv instead of Anaconda (I use uv in my other course and it's becoming the new standard in Data Science, used for example with MCP) - it's refreshingly simple and powerful to use, and the number of environment issues has been far less with uv. But it's not without issues; it comes with a couple of problems of its own (permissions issues on some systems during the install; PATH variable issues; ...).

Anyways, I'm sorry for the long answer and I hope this makes sense to you. Generally I'm super sensitive to merging anything outside "community-contributions" folders - I put a lot of focus into trying to keep the repo stable and bulletproof for current students, and it's too hard for me to change the main codebase or approach and stay on top of all the testing..

Thank you once again for the contribution
Ed

bvsatyaram:
Hi Ed,

Thank you so much for your detailed and thoughtful reply. I really appreciate you taking the time to explain the reasoning behind not merging the PR, and for the openness and transparency in sharing the context.

Your explanation makes complete sense, and I totally understand the challenges of maintaining a stable and consistent experience across such a wide range of systems and learners. Given the scale and the support load involved, it’s absolutely the right decision to keep things simple and battle-tested for everyone.

I'm glad the work was useful to review, and I completely agree—it’s best not to merge the PR. I’ll keep experimenting on my own fork, and I’m grateful for the opportunity to contribute and learn through the process.

Thanks again for all the incredible effort you put into the course and for being so generous with your time and insight.

Best regards,
BV

ed-donner:
You bet @bvsatyaram - thanks for your understanding - and Mamba is definitely really great. Thanks again for all your work on this.
Ed

PR #455, opened 2025-06-15
Fix to load Open ai creds for ai technician assignment

kartik6:
Hi there,
I had completed the week1 day 5 assignment and noticed that others had also done it in a really good way.While going through the community solutions, I stumbled upon one solution that truly stood out. I loved how modular and clean it was, especially the clever way it handled setup and avoided repeating code with its exception handling, seriously top notch work!

I was testing it out with an OpenAI GPT model, and that's when I hit an error that it could not find the OPENAI_API_KEY. So, I went ahead and made a few tweaks to get it working seamlessly with OpenAI.

**What's Changed**
I've made the necessary adjustments to ensure the OpenAI API key is properly recognized and utilized from the env, allowing the solution to run smoothly with OpenAI GPT models. I also cleared the outputs but let me know if you would like me to revert that!

### DIFF
--- a/week1/community-contributions/week1 EXERCISE_AI_techician.ipynb
+++ b/week1/community-contributions/week1 EXERCISE_AI_techician.ipynb
@@ -2,7 +2,7 @@
  "cells": [
   {
    "cell_type": "markdown",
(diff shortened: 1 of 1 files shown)


ed-donner:
Thanks @kartik6 , makes sense

PR #459, opened 2025-06-17
llm battle arrena between three llm

MiR-stack:


ed-donner:
hey @MiR-stack thank you so much.. very fun..
I see you still have outputs in there, and the first few characters of your google key is in there. It's probably not a big deal, but it might be better to save this with outputs cleared before I merge? And out of abundance of caution you might want to change your gemini key? Let me know..

MiR-stack:
Hi @ed-donner 

Thanks for catching that! I’ve just cleared all the output cells and pushed the changes. I’ve also rotated my Gemini (and Google) API keys out of an abundance of caution.

Please let me know if there’s anything else I should adjust before you merge.

ed-donner:
That's perfect! Thanks so much @MiR-stack 
Great contribution.
Ed

PR #468, opened 2025-06-21
feat():Using Historical conversation for Multishot prompting

Guru-Deep-Singh:
This PR build on top of day3 notebook. It allows the model to use the knowledge from the historical conversation to answer the question of the user.

Screenshot from 2025-06-21 18-34-06


ed-donner:
Thanks again @Guru-Deep-Singh - thanks for sharing these improvements

PR #469, opened 2025-06-21
simple RAG use cases (csv/logs/personal data)

moawiah:


ed-donner:
Thanks for putting all this together @moawiah and the clear write-up. Great contribution.

PR #475, opened 2025-06-24
Sm branch wk4

bluebells1:
Two projects from week 4:

1. A unit test generator which creates unit tests from python and runs them (in a Gradio UI).
2. A password generator in Gradio.


ed-donner:
wow @bluebells1 a whole set of great projects - thank you so much for the contribution

PR #479, opened 2025-06-25
Week5 Final Project: RAG solution with multi doctype KB and batch embeddings

bluebells1:
This is the project from week 5. The intention was to create a vector db of my own files (from an external drive) which can be used in a RAG solution.

- This includes a number of file types (docx, pdf, txt, epub...) and includes the ability to exclude folders.
- With the OpenAI embeddings API limit of 300k tokens, it was also necessary to create a batch embeddings process so that there were multiple requests.  This was based on estimating the tokens with a text to token rate of 1:4, however it wasn't perfect and one of the batches still exceeded the 300k limit when running.
- I found that the responses from the llm were terrible in the end!  I tried playing about with chunk sizes and the minimum # of chunks by llangchain and it did improve but was not fantastic.  I also ensured the metadata was sent with each chunk to help.
- This really highlighted the real world challenges of implementing RAG!


ed-donner:
Hey @bluebells1 what an outstanding project.
Yes using the batch API is the way to go. Very nice.
My suggestion if you wanted to get better answers would be to pull LangChain out of there - just write the prompts yourself and you'll get much better performance...
But congrats on a significant project and thank you so much for sharing
Ed

PR #486, opened 2025-06-28
Exersize solution  for week 1 day 2

prashantbhat84:
Hi @ed-donner  ,
Here is my solution for day2 week1. Made tweaks to the way the response object is extracted as response['messages'] was not working and it was giving Attribute_Error


ed-donner:
Thanks @prashantbhat84 - it looks like you've changed files in the main repo directories. Would you be OK to move this into community-contributions? Thanks so much
Ed

PR #493, opened 2025-07-03
i made an brochure generator with ollama and gradio

ArmanEsmaeili:
on a mission...


ed-donner:
Thanks @ArmanEsmaeili -  it looks like you've changed files in the main repo directories. Would you be OK to move this into community-contributions? Thanks so much
Ed

PR #496, opened 2025-07-03
Added gmail RAG example for week 5

amoulin55:


ed-donner:
Oh wow that's really great @amoulin55 - I know the gmail API is not easy to handle, so I'm very happy to see this! Thanks for the valuable contribution.

PR #498, opened 2025-07-04
Added my political bias detection project

ahmadkarami:
This project aims to collect the news from multiple media outlets and compare the language, content, and tone to:

Analyze the political bias of each source
Help users identify neutral and reliable news providers
I extract and analyze news from various media (left-wing, right-wing, and neutral), and provide users with a summary comparison of how each source reports the same event.


ed-donner:
Hey @ahmadkarami this is such a great project and I'm really grateful for your contribution.
Hey I'm slightly concerned that Git Submodules might be too complex for some students and might be a support challenge. How would you feel about doing this differently. Instead of including a submodule, include a single Jupyter Lab with a short description of your project, and then an explicit link to your repo, in the Markdown cells of the Jupyter Lab? That would draw attention to your contribution, and might be easier for those new to git. Thanks so much

PR #502, opened 2025-07-08
Week 5 - RAG

shayed-s:
Created a RAG pipeline that uses documents based on transcripts that were pulled from episodes of the podcast The Huberman Lab on Youtube. The documents that are already in the RAG pipeline are mainly about sleep, nutrition and exercise and based on conversations between the podcast host, the neurobioligist Dr. Andrew Huberman and leading experts, such as Dr. Matthew Walker and Dr. Andy Galpin.

There's probably a way to scrape youtube to get the urls of all episodes. That way you could create a chatbot that lets you chat with Huberman and the guests of this phenomenal podcast about all kind of health, science, performance and wellbeing subjects. 

I was wondering if it would be better to use the raw transcripts or create structured documents. There are pros and cons for both options.

I also imagine a version of this might be useful where you create a RAG pipeline based on EPUB files from books you've read.  You could have the LLM quiz you for better retention or use it for future reference.


ed-donner:
Hey @shayed-s this is really awesome. And yes, that would be amazing.
So I'm a bit cautious about merging this in, because there's a lot of content there that might be copywrited, and at the very least it's other people's content. There's a lot of sensitivity right now about using other people's content in chatbots. What do you think - is this in the public domain?
Ed

shayed-s:
Thank you for comments @ed-donner, didn't think about this when I set this up. I don't think is public domain. On their website they state you can't use their content to create a commercial product. I did not intend this to be a commercial product, merely for personal use, but I think it's better to not merge this. 

Do you have any suggestions for improvements on the methods used or the way it's set up?

ed-donner:
It looked excellent from a quick review @shayed-s - thank you so much - if nothing else, I look forward to trying this out myself sometime! If you wanted to add this to the repo sometime, you could update the PR to remove the files, and include instead instructions for others to download them if they wished to add these files (or anything else) to their RAG..
Thank you again for sharing this!

PR #503, opened 2025-07-09
Handling multiple tool calls using the for loop.

aaron-official:
Well i hope this helps but i using a for loop to handle mltiple tool calls is way simpler👍


ed-donner:
@aaron-official Thanks so much for this. It looks like you’ve made changes to the main files in the main directories - was that intentional? Would it be possible to move your changes into the community contributions folder, and leave the main repo unchanged by the PR..
Thanks again for sharing your contribution
Ed

PR #509, opened 2025-07-10
normal code to commented code for better understanding

MiR-stack:


ed-donner:
That's simple & effective @MiR-stack - excellent! A very nice twist on this project.

PR #513, opened 2025-07-11
Feature/multi language code

aaron-official:
Add multi-language code converter with runtime execution

- Extends base converter to support 5 target languages (C++, Rust, Go, Java, Assembly)
- Adds runtime measurement and execution for both source and target code
- Includes Gradio UI with GPT/Gemini model selection
- Features auto-header detection and compilation error suggestions


ed-donner:
Hey @aaron-official - same as my last comment - looks like this is changing the main repo files..

PR #517, opened 2025-07-13
Upload week5 exercise

Zhufeng-Qiu:
Personal Knowledge Assistant with local files, Gmail, Outlook and Google Workspace files


ed-donner:
Hey this looks incredible @Zhufeng-Qiu - hey is the credentials.json OK to be in here? It looks like you only have your client id and not the secret, which I guess is OK, but it looks slightly worrying - let me know if ok!
Ed

Zhufeng-Qiu:
> Hey this looks incredible @Zhufeng-Qiu - hey is the credentials.json OK to be in here? It looks like you only have your client id and not the secret, which I guess is OK, but it looks slightly worrying - let me know if ok! Ed

I removed the secret manually, and it's totally fine! I kept the credentials.json file to show the structure of my project, so that fellow students can better understand my code and reproduce it.

Anyway, thanks for the reminder, and your course is amazing!

ed-donner:
Fabulous @Zhufeng-Qiu merging with huge thanks - what a big project

PR #529, opened 2025-07-18

johnsunam:
I have added 3 way conversation between three friends using gemini, openai and claude.


ed-donner:
A very nice 3-way convo example - thank you @johnsunam !
Ed

PR #536, opened 2025-07-22
Creating a RAG Database Using OCR, (Cleared Outputs)

MuhammadDevX:
I have added a .ipynb file for creating a RAG Database using OCR, I had cleared out all outputs and only this file is what is added nothing else. It is in the community contributions folder for week 5


ed-donner:
Fantastic @MuhammadDevX thank you so much for building this and sharing it with others

PR #546, opened 2025-07-29
week1 contribution - Youtube Podcast Summarizer using Transcript API

7haz:
week1 contribution - Youtube Podcast Summarizer using YT Transcript API no key needed.


ed-donner:
That's great @7haz - a valuable and interesting contribution. Thank you.

PR #563, opened 2025-08-03
Add UV support for setup

rubijn:
Use uv (extremely fast Python package and project manager, written in Rust) to setup project on linux & mac

use pyproject.toml for modern dependencies management

### DIFF
--- a/SETUP-linux.md
+++ b/SETUP-linux.md
@@ -108,6 +108,35 @@ Run: `python -m pip install --upgrade pip` followed by `pip install -r requireme
 If issues occur, try the fallback:
 `pip install --retries 5 --timeout 15 --no-cache-dir --force-reinstall -r requirements.txt`
 
+### Part 2C - Alternative to Part 2 if Anaconda. Use modern uv tool (50x faster to install)
+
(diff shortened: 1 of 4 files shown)


ed-donner:
Hey - thanks for this @rubijn - I love uv and I will add it to the repo in time, with videos. But here's the thing: with over 100,000 people on the course, I have to be super cautious about changes to the setup instructions. I get perhaps 100 questions a day related to environment. Stable though uv is, there are troubleshooting problems that need to be added. Typically with uv, you wouldn't activate the environment directly, and the recommended command to launch jupyter would be `uv run --with jupyter jupyter lab`
Then there are some things that need to change in the labs (like using `uv run`) 
And there would need to be PC instructions.. and I'd need to test the labs across all major platforms with the uv environment..
So I'm totally up for doing this, and I'm grateful for your contribution, but just want to explain why it's probably a bigger deal than merging a PR..
I should mention, over the last 6 months I've also had PRs for Pyenv, Poetry, Mamba and some others - and I've not merged them - they each come with added complexity and drawbacks. Definitely uv is my favorite..
Hope that makes sense and thanks for sharing
Ed

PR #565, opened 2025-08-04
Mb llm excercise

mbhoge:


mbhoge:
Flow structure updated at the start of the code

PR #601, opened 2025-08-16
Proposed UV toml working with Mac on Intel

gilesgamon:
Having come back from your latter AI course, this seems like it would be handy and likely innocuous but perhaps include a brief additional notes resource?


ed-donner:
Hey @gilesgamon - yes I'll add this with full instructions and rerecord the intro soon.. thanks for sharing

PR #607, opened 2025-08-18
Added Exercise for Week3.

hafizabrarshah:


ed-donner:
Thanks for making this valuable contribution and sharing your work with us all.
Ed

PR #627, opened 2025-08-26
email generator subject and body from given topic

ravirch:
Hello ED, Added auto email body generator and subject generator using Ollama from week1.


ed-donner:
Great - many thanks

PR #628, opened 2025-08-26
Suggested extension: auto-correct city name typos in tool calls from Airline assistant

dpipchenko:
I suggest to demonstrate that tools can do more than just passing in parameters. They can prepare them if needed.

In our case, if a user types "Tokio", then with my small extension it will be automatically fixed to "Tokyo" so that the price is found. Otherwise OpenAI says "I don't have price information for Tokyo at the moment" auto-correcting the obvious typo just in its output but not in the preceding tools call.


ed-donner:
Hey - would you be OK to put this as an improvement in community-contributions? I do appreciate it's an improvement, but I prefer not to change the main repo except where absolutely necessary to avoid disrupting students..

PR #629, opened 2025-08-26
Add week4 challenge to community-contributions

Oluwaseyi-A:
This notebook contribution is one of the week 4 challenges: a code generator tool that writes unit test cases. When run, the user is able to upload a python file, and a Frontier model spurns out the code for unit testing. It also has the capability of running the unit tests. The approach is fully cross-platform — Windows, macOS, Linux — because nothing in this notebook relies on OS-specific behavior.


ed-donner:
Terrific - an important use case - thank you so much for sharing your work!

PR #632, opened 2025-08-28
Week3Day5 - Dataset generator

prateektambe:
This PR is for Week3Day5 - Dataset generator.


ed-donner:
Great - thanks for this valuable contribution - dataset generation is an important use case

PR #639, opened 2025-09-01
Python C extension generator using OpenAI GPT-5 

Carbaz:
This branch implements an alternative way to generate high performance C code from Python code with the heal of a frontier model.

OpenAI's GPT has been chosen for it's built in capability to enforces a JSON schema that the output will "always" match.", although it should be really simple to move to other model just changing the client instance and optimization request according to the API.

It uses CPython feature to import and run C compiled code inside Python scripts so it allows developers to re-implement some functions from their applications using C and compile them into libraries/modules that can be imported from python code.

That way Python's high level code can be combined with C high performance when required without the need to migrate the whole application/service to C, just the functions and classes requiring extra optimization would be rewrote.

Model will provide not only the C code but the compiling script and a performance comparison test script.

Notebook provides examples of use, code cells to test and a Gradio app that includes all the steps from the original python code to testing the generated C extension module.

C compiled modules will be left on the local folder with a `.pyd` (for windows) or `.so` (for UNIX) extension.

To reuse the compiled modules on other projects just copy it to the projects folder, the python interpreter site-packages, or somewhere on the python path, and then import it as usual.

But mind that as they are C compiled they are specific to the os and architecture they where compiled for.

The folder includes some examples of generated code and a README that describes the Gradio app interface.

> [!NOTE]
> 
> Code has been tested on Windows only, the compile step, and later module usage, should be tested on Mac or Linux too.
> There is no need to test with complex code, the Gradio's app default code should work to validate the compile chain.
>
> Also, as on the original notebook, the C compiling tool-chain must be installed on the machine.


ed-donner:
wow this is an amazing contribution @Carbaz - congratulations and thank you for sharing this

Carbaz:
Thanks you for the course, it's been impressive.

PR #653, opened 2025-09-09
assignment: week2 exercise final

Lythmass:
Combining streaming and tools, I built a coding tutor AI that can take inputs from voice and generate image if asked about calculations.


ed-donner:
Great! Thanks for making this valuable contribution and sharing your work with us all.
Ed

PR #655, opened 2025-09-09
Added my contribution for day 4

bharathchandrareddy:
The code given by tutor in the jupyter notebook throws an error when more that one city name is entered. I've made changes in the function tools to handle multi city input.


ed-donner:
Very nice improvement - many thanks!

PR #657, opened 2025-09-11
Murtaza Lodgher - Day 1 contribution

malodgher:
Create a small app that generates a song that has to include user-inputted keywords. User can input music genres to to tell the AI to write in certain genres of music. User can also input any artists that the AI can familiarize with.


ed-donner:
That's so great! Very nicely done. Thank you for sharing.

PR #666, opened 2025-09-16
integration of OpenAI-based page summarization

vinitparak:


ed-donner:
Very nice - thanks for the contribution

PR #669, opened 2025-09-19
Week 4 excercises: added Gemini and Python Code Documentation Assistant

Krabulek:


ed-donner:
Wow there's some meaty projects here - thank you so much

PR #672, opened 2025-09-20
Ruby RAG example with a console chat app added

dmbavykin:
# Ruby RAG console chat app

This mini app shows how similar result can be achieved using Ruby language.
Please, check README to find the discovered insights and challenges. Also, you can find there some instructions how to run the app. 

For testing the approach another fake test data was generated which can be found at `knowledge_base` folder.


ed-donner:
Wow this is a big project! Exciting to get some ruby into the repo.. thank you!

PR #678, opened 2025-09-21
Simulate the function which reserve ticket by mock api and make llm use it as tool.

tylin7111095022:
1.Use llm (openai API) to simulate the function which reserve ticket by mock api.
2. Week2 Exersice
3. Use llm for python code acceleration (Week 4 exercise)


ed-donner:
Great! Nicely done.. thank you for sharing both!

PR #683, opened 2025-09-23
Add batch mode for processing multiple URLs 

irytck:
This PR introduces a new batch mode that allows processing multiple URLs in a single run and returning centralized summaries.


ed-donner:
terrific! thank you for sharing

PR #685, opened 2025-09-24
Shabsi4u brochure with UI

shabsi4u:
Uploading a python version of LLM brochure generator with UI - that supports 3 frontier LLM providers


ed-donner:
And 18,000 lines for this one..

shabsi4u:
Fixed, sorry again!

ed-donner:
hmm it's better not to include a uv.lock because this is still 5,500 loc..

PR #689, opened 2025-09-26
3-way conversation with representatives from night city, discussing dog town

mfoland14:
This is a fun exploration of how much impact the system prompt has over the generated responses of the AIs.  Initially I did not include the round robin aspect which led to some interesting but disjoint conversations.  I also added the note to response to each others points which helped to create even more continuity. 

Included two example conversations with very minor changes to the system prompt which in my opinion helped get the conversation started properly.


ed-donner:
Very fun!

PR #691, opened 2025-09-27
Week 5 improvement

mihaly-hazag:
On Windows the original code could not read the directory. The change aims to fix it

```
import platform
....

"if platform.system() == 'Windows':\n",
    "    folders = glob.glob(\"week5\\knowledge-base\\*\")\n",
    "else:\n",
    "    folders = glob.glob(\"knowledge-base/*\")\n",
    "\n",
```

### DIFF
--- a/week5/community-contributions/day2_windows_fix.ipynb
+++ b/week5/community-contributions/day2_windows_fix.ipynb
@@ -1,0 +1,196 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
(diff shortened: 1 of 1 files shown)


ed-donner:
well that's weird. glob.glob is actually supposed to be platform independent and handle either type of slash, and I've tested this on windows (and you'll notice, there are 1 or 2 other students on this course 😁) 
Is there anything special about your setup?

mihaly-hazag:
you are right, I used it from VSCode connecting to the Jupyter lab, it is a slightly different setting

PR #696, opened 2025-09-29
Update SETUP-mac.md (Minor Typo)

rakeshkumardas991:


ed-donner:
hmm. actually "Science" is sometimes used instead of "Data Science", rather like "Engineering Lead" and "Software Engineering Lead"

PR #701, opened 2025-10-01
Fix environment.yml pytorch install with Cuda

maufonfa:
The Pytorch installation using conda was giving the following error:
Downloading and Extracting Packages: InvalidArchiveError('Error with archive C:\\Users\\my_local_user\\anaconda3\\pkgs\\pytorch-2.8.0-cuda128_mkl_py311_hd9a8a8a_300')
I found that adding pytorch to the channels, and adding torchvision and torchaudio fixed the problem.
Hopefully this sounds like a good solution

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -1,5 +1,6 @@
 name: llms
 channels:
+  - pytorch
   - conda-forge
   - defaults
 dependencies:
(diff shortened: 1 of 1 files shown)


ed-donner:
The proper solution to the root cause is to enable long file names on Windows, as documented in the Setup guide..

PR #714, opened 2025-10-06
feat: Final submission for week 3 exercise synthetic dataset generator

Mr-Tson:
Created a gradio app with Llama 3.1-8B-Instruct, where the user can choose between 3 different dataset types:
1. Instruction-Response Pairs
2. Multi-Turn Chat Dialogue
3. Text Classification"
Based on that, a dataset will be streamed to a text ouput and the user can download the dataset as a csv-file.


ed-donner:
hmm something seems up with this file.. it's large and I can't see it in github?

Mr-Tson:
Thanks for the feedback! I have cleared the cell outputs and updated the PR to include only the Week3 Exercise file. The notebook should now render properly in github.

ed-donner:
thanks so much!

PR #715, opened 2025-10-06
Add my challenge solution of week2 Trip planner 

dharan-070404:
Created chatbot using Tools for ticket price and flight booking,multimodal features with Gradio interface using OpenAI api


ed-donner:
Great! Thanks for making this valuable contribution and sharing your work with us all.
Ed

PR #734, opened 2025-10-17
add solutions for week1

NatKacz99:
I would like to send two projects: 
One is a simulation of an accommodation assistant that uses the hotels.co website, and the other is a company brochure


ed-donner:
Great! Thanks so much - very nice

PR #737, opened 2025-10-18
Bootcamp (week1): exercise solution - MathXpert

ranskills:
# Bootcamp: Week 1
## A Math Expert that throughly explains solutions step-by-step.

- Completely uses free cloud LLM providers, i.e. OpenRouter and Ollama Cloud, so user only have to bring their keys.
- Allows for dynamic selection of an LLM provider depending on provision of API Key for the provider


ed-donner:
Terrific - thanks so much @ranskills

PR #762, opened 2025-10-20
Wellness companion

abdoulrasheed:


ed-donner:
Thanks @Abdoulrasheed - much appreciated, and the wellness app is a really cool idea. For future PRs, please also include a screenshot and a key takeaway in the PR comments - thanks so much

PR #766, opened 2025-10-21
Javi Bootcamp PR Week 2

javieroteromarquez:
Main focus was Streaming work with func tools (not fun at all!).
Then I simply stitch up a few pieces together (several models, audio output) and created an LLM_as_tool (a bit hardcoded, but nice).
So the workflow for the tool is make a question, you get an explanation, and if you send code back the LLM_as_tool checks it with a different model, system prompts and outputs different template.
image
image


ed-donner:
Wow that is really terrific @javieroteromarquez . Goodness that streaming code looks no fun at all. You're gonna kill me after the next course.. streaming with tools is super-easy when you use OpenAI Agents SDK. Anyways excellent project, thank you for doing this and sharing great work.

PR #771, opened 2025-10-21
Synthetic data writer

iamumarjaved:
Umar - Bootcampt

So as per instructions I have created the synthetic dataset generator which creates datasets from llama and openai and we can define number of records to create and at the end it creates downloadable file and with that it also shows the preview of file displaying data in it and you can also select data from specific country or global.

dataset
preview


ed-donner:
very nice - thanks Umar - great

PR #775, opened 2025-10-21
add nbformat dependency to pyproject.toml

escalderong:
nbformat is required for plotly's graphs.


escalderong:
Most people would not encounter this problem, but if you are on a fresh computer (like just happened to me) you won't have that installed around

ed-donner:
Muchas gracias! It's very rare that I merge changes outside community contributions, but this is a very clear case! Sorry for missing it, and thank you for the fix!

PR #778, opened 2025-10-21
Javi Bootcamp Exercise Week 3 and fix folder week2

javieroteromarquez:
Includes small Gradio App in Google Colab that uses a couple of local models through HF to create flashcards that are displayed in markdown
telegram-cloud-photo-size-4-6035361974071266443-y

As an insight, I had to postprocess a bit the tokens output playing with regex to properly display (eg, deleting the template before using regex to only display flashcards). I guess this is another bit that frontier apps do for us when we just call them!

PS: Sorry I messed with the file location on week2! Just fixed it


ed-donner:
oh wow that's a really creative idea and project - super nice Javi
Ed

PR #790, opened 2025-10-22
stock code generator

iamumarjaved:
Umar - Bootcampt

So I have on the **Trading Strategy Code Generator** in which the program does the following:
- creates code based on the strategy.
- user can select the default strategies or suggest and based on that it will generate python code.
- User can run directly in the browser and check the code and it's results

Screenshot 2025-10-22 at 18 33 54


ed-donner:
Oh wow - that's really nice. Great job, Umar
Ed

PR #791, opened 2025-10-22
Community contributions branch

Godbero:
Way to summarize all websites (JavaScript too) without Selenium, using jina.ai to turn website content into Markdown before sending to LLMs.


ed-donner:
Thanks for making this valuable contribution and sharing your work with us all.
Ed

PR #792, opened 2025-10-22
Javi Bootcamp Week4 exercise: Code and test app

javieroteromarquez:
Create small Gradio app that includes both, how to add comments (at diff levels of expertise) and create unit tests.
Played with Gradio, tuned it to look like VSCode, and my insight was to learn how to sync the input element so it was shared.
image
I shared the app with a friend (with gradio share) and he was really impressed and was using it for his code until I reminded him it was on my API!!


ed-donner:
I love it Javi! Yes isn't gradio great when you get the hang of those callbacks.. very nice

PR #798, opened 2025-10-23
Community contributions fernando

nandoclal:
About the exercises It´s really It is very easy to leave colleagues amazed with the message-response (the magic behind a particular solution are the prompts). And be aware of the APIs for several services is very useful, so that we can prepare solutions that explote the strengths of each model.
Day 2
image

Week1 Exercise
image


ed-donner:
Thanks Fernando! Hey it looks like you didn't clear the outputs of the notebook so this would be sending full outputs. Please could you update the PR with the labs with the outputs cleared? Many thanks
Ed

nandoclal:
Thank you Ed.

Correction in progress.

PR #801, opened 2025-10-23
[Bootcamp] - Tochi - Week 4 Exercise - Python to TypeScript Code Converter

Tochi-Nwachukwu:
# [Bootcamp] - Tochi - Week 4 Exercise - Python to TypeScript Code Converter

**Solution Snippets**
image
image

**Key Takeaways.**

- I learned about using Gradio to display code, both as input and output.
- I learnt about using LLMs to convert from one code to another.
- However, I noticed that the code conversion took too long for gpt 5 Nano, and a few times, the code has errors when it is run. Other models should be able to produce better output


ed-donner:
Great example and interesting observations. Nicely done, Tochi

PR #802, opened 2025-10-23
[Bootcamp] - Nikhil - Week 2 Exercise - Agentic AI Real .com Domain Finder

rnik12:
Screenshot 2025-10-23 at 1 52 59 PM

### **Utility**

* Finds a list of available `.com` domains based on **industry**, **target customers**, and **domain description**.
* Uses free **RDAP GET requests** and an **LLM tool call function** to check domain availability.
* Refines preferences and results through **multi-turn prompting** — users can reply to refine or update the domain list interactively.

### **Key Takeaways**

* Learned how to build an **agentic AI tool** with a **Gradio frontend** - handling multiple LLM calls was initially intimidating.
* **Cost control** is crucial in agentic AI systems; significant effort is needed to manage and optimize it.
* It’s best to ensure that **tool call functions** are either owned by the company or have **SLA guarantees**, since any issues in these functions directly affect agent performance and lead to a poor user experience.

### **Notes**
* The RDAP domain availability API is slow, please let the tool run for atleast 400 seconds before killing it. Logs are visible in Notebook to check where it is stuck.


ed-donner:
Nice, thanks Nikhil - interesting example.

PR #806, opened 2025-10-23
(Oct 2025 Bootcamp): Add RoboCare AI Assistant project files including database, README, a…

hopeogbons:
# RoboCare Assistant

RoboCare AI Assistance

## Why I Built This

While working on a caregiver matching platform for **MyWoosah Inc** in the US, I faced a real challenge: how do you efficiently match families with the right caregivers when everyone has different needs?

Families would ask things like:
- *"I need someone for my mom on Monday mornings who speaks Spanish"*
- *"Can you find elder care in Boston under $30/hour with CPR certification?"*

Writing individual SQL queries for every combination of filters was exhausting and error-prone. I knew there had to be a better way.

That's when I discovered the **Andela LLM Engineering program**. I saw an opportunity to transform this problem into a solution using AI. Instead of rigid queries, what if families could just... talk? And the AI would understand, search, and recommend?

This project is my answer. It's not just an exercise—it's solving a real problem I encountered in the field.


ed-donner:
Nicely done, Hope - and I love that you're applying this to your real world projects.. great!

PR #811, opened 2025-10-23
Week 4 : Using LLMs to translate Python Code to C++ compiled code

TheTopDeveloper:
Here I am using GPT, Claude and Gemini to take as input Python code and then create a compiled C++ version. 
The key takeaway was that the formats of output by the different LLMs differ and to one has to be keen not generate a one for all solution but take time and monitor how the LLMs approach the questions


ed-donner:
Very nice Joshua! Just a reminder to include your note in the description as I can't always remember who has which github username.. also a screenshot in the PR comments in the future please. The sorting algorithms is a nice touch

PR #816, opened 2025-10-24
Add week 2, day 1, N way conversation solution

davebelliveau:
Implements SNL Coffee Talk character conversation using multiple models. Supports OpenRouter, OpenAI, and Ollama with automatic fallback. Characters and topics are dynamically generated by the base model.


davebelliveau:
Removed outputs and force pushed.

ed-donner:
Very nicely done - thank you!

PR #819, opened 2025-10-24
Assignment : Week 5 by Bharat Puri

bharat109puri:
Hi @ed-donner 

I have fixed all issues of langchain and update the code changes and validated Vector Graph creation.
I will also create another PR for Gmail integration with LLM.

Pasted Graphic 16

Pasted Graphic 17

image

Take Away:
1. Get better understanding of Langchain and its respective method, while updating the methods.
2. Its one of the interesting feature of AI, which i am going to implement for my personal Knowledge Base.


ed-donner:
Nice! You'll see in the team project that I've found a way to do this without needing to use the 'classic' package at all, that also removes the memory stuff..

PR #827, opened 2025-10-24
Added Exercise week6

javieroteromarquez:
A bit of a disappointment here, I don't think I managed to find the click that would improve the situation.
I've tried more training, less training, more data, more prompting, and all of them were a bit worse than default.
- More training: Trained too fast (2 epochs, 16 batch size LR 0.1), not enough convergence resulting in large error ($153)
image

- Prompting on inference only: Reduced error down to $88
image

- Prompting training/inference with slow LR (0.01): Same error ($89)
image

- Nicely converged model with LR 1 yielded same error ($87)
image


ed-donner:
Very nice experiments tho! Week 7 will save the day. I'm gonna try this myself over the next week to see if I can find anything.. Thank you for sharing

PR #829, opened 2025-10-24
Week 2 Task: Trending Coins with OpenAI and Llama, using Gradio Chat

sammurira:
Week 2 Task: Trending Coins with OpenAI and Llama, using Gradio Chat


ed-donner:
Looks like you haven't cleared your outputs here @sammurira 

FYI one of the PRs that you raised before had a filename with a colon in it - this is not a good idea as it can break on some Windows platforms. I had to back it out.

ed-donner:
OK thanks much

PR #835, opened 2025-10-25
Add carl-grp's homework for week1 day2 (cleared outputs)

carl-graup:
This is a proposed solution to the homework for day 2. It contains a simple scraping of Apple's newest iPhone websites that are then compared by a local run Ollama LLM and ChatGPT 5 nano. Enjoy the comparison on how well the models perform!


ed-donner:
An ingenious and useful assignment for gpt-5! Thanks @carl-graup

PR #842, opened 2025-10-26
[Bootcamp] - Nikhil - Week 4 Exercise - Multi LLM Testing for Pytest Generattion

rnik12:
Screenshot 2025-10-26 at 8 24 53 PM

### Key Takeaways
* Learned that its easy to integrate other LLM providers, no need to just stick to OpenAI.
* Almost all models even LLama 3.1 8B seems very good at generating pytest for simple functions.


ed-donner:
Really nice Nikhil - great use of the Gradio code widgets. Super.

PR #843, opened 2025-10-26
[Bootcamp] - Nikhil - Week 5 Exercise - RAG on Legal Bare Acts Txt Copied from PDF

rnik12:
Screenshot 2025-10-26 at 9 11 39 PM

### Key Takeaways
* Learned how to do RAG based QNA on our own knowledge base
* RAG is not performing good with MiniLM having 256 token limit. A larger token limit embedder will be good.
* Overlap Chunking is better chunking strategy than just non overlapping chunking.
* Source data need annotations, direct QnA with RAG is going to fail as matching chunks do not tell enough information.


ed-donner:
Another fine project in your domain Nikhil! And using data from your data generator.. wonderful. Ahh those observations are good, and one of them will give you a very quick win with the group project

PR #844, opened 2025-10-26
[Bootcamp] - Nikhil - Week 6 Exercise - Finetuning gpt4o-mini

rnik12:
Screenshot 2025-10-27 at 12 38 08 AM
Screenshot 2025-10-27 at 12 36 35 AM
output

### Key Takeaways
* Learned how to fine tune frontier models. My finetuning is performing worse than the course probably due to random seed used during finetuning. Switching to another model may bring better results.
* Found that it is not that easy to change anything as change is bottlenecked by training data set. We need to strategies first how to ensure accuracy from LLM for example we need CoT data for improving price prediction but we do not have CoT in our training dataset.


ed-donner:
hmm interesting.. perhaps you could generate enriched data with a category, brand, quality, then price bracket, then the price - that would force the model into a CoT style mode. That's a nice idea!

PR #853, opened 2025-10-27
sach91 bootcamp week3 exercise

sach91:
Sachin's pull request for week 3 exercise in bootcamp.

A travel agent based on HuggingFace's Llama model exhibiting biased to one particular destination, also provides translation support using another model call.

Some insights:
1. I found it tough to make huggingface APIs work on local system, there were issues with model downloading in jupyter.
2. Working on google colab with huggingface is easier.

Below is a snapshot of the gradio app that provides translation using a second call. It was somewhat interesting to maintain both English text and translated text in the app, but use only original English text for sending history to LLM. 
 
image


ed-donner:
OK many thanks

PR #854, opened 2025-10-27
sach91 bootcamp week4 exercise

sach91:
Sachin's pull request for week 4 exercise in bootcamp.

**A HuggingFace LLAMA code generator and validator, using gradio interface.**

Some insights:
1. Smaller open source models (Llama3.1-1Billion) are also able to generate decent code.
2. This small model also did a good job for code documentation.

image


ed-donner:
Hey Sachin - the idea is for you to make a tool to take existing code and add test cases or comments.. this seems to be a ChatBot that you're asking questions to. Am I misunderstanding?

sach91:
To simulate having an existing code, I make first call to the LLM to give me a code, strictly without any comment or documentation.
Then I use this code to add comments using another call to LLM. 
So, the second call does the expected. The first one provides a simulation environment. 
If I need to keep only the second part in this exercise, or improve on something, please suggest.

ed-donner:
OK I see - fair enough - thanks @sach91

PR #855, opened 2025-10-27
[Bootcamp] - Nikhil - Week 7 Exercise - Finetuning Llama-3.1-8B

rnik12:
finetuned
Screenshot 2025-10-27 at 3 04 20 PM

### Key Takeaways
* Learned how to fine tune open models like Llama-3.1
* Learned that finetuning open models is better than frontier models as there is no gating or alignment checks.
* Finetuning on Google Colab is very slow, I reduced training data sets to 1K from 100K.
* Performance maybe improved if we spent trained on entire 100K dataset and using bigger and faster GPUs.

### HuggingFace
* Link to finetuned model uplaoded on hugging face https://huggingface.co/nikhilr12/pricer-2025-10-27_07.15.03


ed-donner:
Hey - thanks so much - hmm there seems to be 8,000 lines in these files. Perhaps just check in a link to the colab in a short file instead? Did you try any hyperparameters different to the course? Thanks so much

rnik12:
Hi @ed-donner, I tried to clear out the output logs and now the lines are decreased from 8K to 6K but still very huge. I only changed few code blocks compared to original. Can you suggest what can be done here ?

ed-donner:
hey - how about just check in 1 notebook that has a short explanation, and has a link to the colab that you have enabled sharing, so that I can follow the link and see your colab in google?

rnik12:
Hi @ed-donner  Google Colab is not allowing further finetuning and is rejecting connections to GPU instance.

However I have updated links to latest colab. The test notebook points to earleir finetuned model only. Let me know if it works.

rnik12:
@ed-donner latest PR has only 6 LOC changes.

ed-donner:
Ahh yes that's a nice way to do it - thanks so much
Ed

PR #856, opened 2025-10-27
sach91 bootcamp week5 exercise

sach91:
Sachin's pull request for week 4 exercise in bootcamp.

**RAG based gradio solution to give information from related documents, using `Llama3.2` and `nomic-embed-text` over Ollama**

Some insights:
1. RAG is an excellent method to dynamically populate the context for an LLM query.
2. SQLite is an easy to create a database for analysis work.

Below is a screenshot of the gradio app which load the documents on button click, and initiates similarity search from SQLite database for RAG, enabling using local Ollama model for tokenization and query. 

image


ed-donner:
Hmm. I'm not clear on why you did this - rather than using a vector datastore, you appear to have manually created SQLLite tables, and be reading them in and calculating cosine similarity manually? What was your goal here?

sach91:
If the vector db is unavailable, I create it the first time with required syntax, for later population. 
When documents are provided, I create the embeddings and upload them to the vector database. You can check below line from where I populate the documents from knowledge base.
    _folders = glob.glob("../../week5/knowledge-base/*")_
Later on LLM query, the similarity search is performed using the cosine formula.
The functions are separated by class, that may have caused the confusion.

ed-donner:
hmm no my confusion is that rather than using a vector database, you appear to have stored all the vectors in SQLLite and then manually calculate the cosine similarity? If that's what you intended then fair enough

PR #859, opened 2025-10-27
Javi Bootcamp Exercise Week 7 using Phi4 unsloth

javieroteromarquez:
Finetuned phi-4-unsloth-bnb-4bit. Due to not fiting in the NVIDIA GPU using transformers, I used library unsloth (I guess I couldn't find the proper quantization way, as that fitted in memory with 10GB).
Did a few retraining to see how it performed, although Colab was stopping and kicking me every once in a while.
Because I was checking the lite dataset, I also performed the evaluation on original llama, ed's finetuned llama, my original phi4 and finetuned at 4k steps and 6k steps (out of 22k training points)
Llama 3 base 35.86$
image
Ed's Llama fine tuned 16.2$
image
Base Phi4 26$
image
Phi4 ft-4k steps 21$
image
Phi4 ft-6k steps 20$
image

Insights: Might be worth paying a few dollars and not fighting with free-limits. Might also be worth to use higher boxes (or smaller models, I was very stubborn on using phi4) to alleviate restrains on how to retrain. Managed to play both with wandb and HF to retrain from midpoints


ed-donner:
Oh this is really cool, Javi. This is with the Lite dataset (Home Appliances) right?

javieroteromarquez:
Yes! I was getting frustrated at the beginning because the model was OOM just by 50 MB so I decided to switch to the lite dataset, and then I just ran everything with that

PR #864, opened 2025-10-27
Javi Bootcamp Exercise week 8 and cleanup week 5 (no outputs)

javieroteromarquez:
I wanted to add something to the random forest and ensemble. That's why I used diff model (catboost, I was going to use category but then on inference that's not being extracted so it was a bigger task) with nice regularization. I've also added a gbt for the ensemble, because when trying ElasticNet with no intercept and positive values most of it went to Frontier model (~80%) and very heavily regularized.
Also combined my specialist agent (phi4 ft on lite dataset) with the rest. In the end, there was no improvement (58.5$ v 54.5$) but I can sleep better :D 
image

### DIFF
--- a/week5/community-contributions/week5_jom/Exercise_week5_jom.ipynb
+++ b/week5/community-contributions/week5_jom/Exercise_week5_jom.ipynb
@@ -47,21 +47,10 @@
   },
   {
    "cell_type": "code",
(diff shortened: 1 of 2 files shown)


ed-donner:
That's all good with me. You're the first to finish!!

PR #866, opened 2025-10-27
Week7_exercise

CosmusMutuku:
****Exercise Description**:**

- I fine-tuned the OPT-125M model (tried many models but i couldn't manage due to resource issues but i finally settled to this)  using LoRA to estimate the fair market price of Amazon products from textual descriptions.
- Used a subset of 2,000 training and 500 evaluation samples from the  dataset was used to keep training lightweight.
- Framed the task as an instruction-following regression problem. Alos, prompted the model to output a single numeric price.
- Goal: exploring parameter-efficient fine-tuning (PEFT) under limited compute and dataset size.

**### Results (after revision)**

- The baseline model (without fine-tuning) achieved a MAE of $50.59 and RMSE of $70.34.
- After fine-tuning with LoRA, the model achieved a MAE of $48.79, RMSE of $122.14, and R² of -0.0026.
- The lower MAE suggests marginal improvement in average error magnitude. Althought the fine-tuned model’s RMSE worsened slightly
- The Results indicate that the model learned some basic price reasoning patterns, but performance remained unstable—likely due to limited data, small model capacity, and noisy targets..

### Experiment's Challenges
- Training large models on Colab led to frequent runtime timeouts (unfortunately after waiting for half a day or so) and GPU memory constraints.
- The workaround was to reduce dataset size, use a smaller model (OPT-125M), and apply LoRA adapters. (performance was not the best but there ws some improvement compared to the baseline results)

Screenshots:
image
image

image


ed-donner:
Hey! I'm not following that last diagram - it looks like the prediction is $0 throughout? What am I misunderstanding?
Ed

CosmusMutuku:
Hi got it.
Now, after re-verifying the datasets and fixing the indexing logic, I confirmed that the true prices were correctly included in the train and evaluation JSONs. Also the evaluation code now extracts them properly.
Of course, the results changed significantly — but still the fine-tuned model outperformed the baseline.
(the previous comment has been updated to refect this. Thanks for the catch)

ed-donner:
hmm I'm not totally understanding these massive error numbers. A Mean Absolute Error of.. $33 million?! I'm misunderstanding something. Your baseline results are off by an average of 33 million dollars? Sorry if I'm being dozy

CosmusMutuku:
Haha no worries — I was also thrown off by that number at first. The big culprit was the baseline model itself — I used OPT-125M, which is a very small language model meant for text generation, not numerical prediction. It’s a pretty dumb baseline for a regression task.
I landed to the model after long iterations to find what would fit in my computational capacity. But will give it a more llok and see

CosmusMutuku:
Hey Ed, as promised I went back and dug into the setup. I think i finally figured out what was going on with those crazy $33M baseline errors.

Basically, the issue came down to how I was handling the model’s numeric outputs and preprocessing:
- I hadn’t properly reversed that with np.expm1() during evaluation, so the predictions ended up being interpreted as huge dollar amounts.
- The regex extraction was too loose.At some points it grabbed stray numbers from the prompt context instead of the model’s actual prediction. Ths in a big part inflated the errors.
- The model’s raw output sometimes included partial text or repeated tokens. I was decoding the entire sequence rather than just the newly generated tokens.
- After those fixes, I got 'better/ealistic results'

The model’s still far from great at price reasoning. But at least  i believe now it’s behaving numerically sensibly.  so the earlier massive errors were just a preprocessing and evaluation mismatch rather than an actual modeling issue, which is a lesson learned now.
I have updated the first comment and hope this now is ready for merging. Many thanks!

ed-donner:
Haha yeah I figured it was in the extraction. Actually none of these models are intended to produce numbers - these are all classification models, not regression.. it's kind of amazing that this price prediction with an LLM works at all. Especially with a 125M model. But yes those crazy numbers were more likely to be a regex issue. Product numbers becoming prices or something. Thanks for tracking it down, and nice experiments..

PR #872, opened 2025-10-28
Domienbakker/day1 solution

domienbakker:
Solution for Day 1 selenium scraper.
This require following extra dependencies:
- Selenium webdriver installed on the user computer.
- Python packages: 
selenium==4.11.2
urllib3==1.26.16

Webdriver only worked together with Python and this exact package versions.
Tested on Windows, but with some luck it will work with MacOSx as well.


ed-donner:
Very nice - thank you

PR #877, opened 2025-10-28
Umar Javed - Bootcamp

iamumarjaved:
Good Morning @ed-donner , 
so do you to storage and time constraints I have used Llama 3.2 1B, as it have lower capacity and I used less data as it takes a lot of test different parameters and train I was able to make the Prediction error to 79$.

image
image
image


ed-donner:
That's excellent! Llama 3.2 actually does very well indeed. Thank you..

PR #880, opened 2025-10-28
GenAI Bootcamp - Week 3 Synthetic Data Generator

smqd19:
This is the PR request for Week 3. My synthetic data generator generates documentation about RAG, langchain. It generates technical documentation like overview, faq. 

Key Takeaways:

- I experimented with different temperature values to see how the output was affected. 
- Initially, I set the temperature to 0.0, the result was very to the point documentation, it was very formal and straightforward.
- I then tried increasing the temperature to 0.6, 0.7 and the output started becoming more intuitive. It no longer followed the pattern defined to the dot, however the LLM added more elements on its own.

Week 3


ed-donner:
Thanks so much - nice

PR #883, opened 2025-10-28
Gen AI Bootcamp - Week 5 Rag

smqd19:
I have created a rag using the synthetic data generated in Week 3. To add more variety in my knowledge base, I generated multiple different type of documents.

Key Takeaways:

- Initially, I was using recursive text splitter for my chunking, however the output of RAG was not that great. It failed on even very simple questions like the "What is RAG?"
- I then switched over to semantic chunking to see if the results would improve. The results actually did end up approving as a result of using semantic chunking.

Week 5


ed-donner:
That's a pro-move! Nicely done. Seems like a good idea for the group project...

PR #884, opened 2025-10-28
Gen AI Bootcamp - Week 6 Fine Tuning Frontier Model

smqd19:
I had tried the training a total of 3 times. The first iteration and second iteration performed badly, especially the second one, it was even worse than the first one. In my third iteration, I tried playing around the system prompt, changing it multiple times to create one on which the performance improved significantly. The loss at the end was $66.47.

Key Takeaways:

- Having a very well written system prompt affects the training quite a lot.

Week 6


ed-donner:
It took me a bit to understand the diagram, but I get it.. nice investigation..

PR #889, opened 2025-10-29
GenAI Bootcamp - Week 7 Fine Tuning Open Source LLM

smqd19:
Initially, I was trying to train Llama-3.1-8b, however my Google Colab session kept on reaching its runtime limit. The training couldn't really be completed. I then shifted my approach. I tried the training locally. 

- I first created the lite pricer dataset for appliances.
- Instead of using the Llama-3.1-8b, I used Llama-3.2-1b. I was finally able to train this then.

Week 7


ed-donner:
Wow that's really awesome, thank you!

PR #891, opened 2025-10-29
Week8 day5

iamumarjaved:
Umar - Bootcamp

HI @ed-donner , 
so in addition to LLM based estimation with RAG. I have added XGBoost pattern matching on embeddings. I started working on Flight-fare usecase I came to this usecase by having the similar one in mind being implemented in videos but the issue was no appropriate dataset was there for this so I created my own.

You mentioned in last videos to create tests and I have done that too,  I didn't go for unittests or any specific framework based testing because the implementation was not of that scale. 

Screenshot 2025-10-29 at 13 01 39
Screenshot 2025-10-29 at 12 51 39
Mobile


ed-donner:
This is really great Umar - I love it! I'm quite jealous.. this must have been a lot of fun. XGBoost is great. Very nice indeed.

PR #895, opened 2025-10-29
kwabena_bootcamp

kbaah:


kbaah:
data generator

I am simply amazed at how hugging face makes it easy to work with these open source models! Hopefully a commercial version of this solution can help to generate some good data for training a model that can be later used to help users customize resumes

PR #896, opened 2025-10-29
Kwabena Bootcamp Week 4

kbaah:
unit test writer

I wanted to restrict max tokens but realised the syntax for doing that with Openai's models differs from other languages so will try that in a different implementation


ed-donner:
Sure, nicely done - good example

PR #897, opened 2025-10-29
GenAI Bootcamp - Week 8 Ensemble Model

smqd19:
I had planned to make changes in the Ensemble Model. My plan was to replace the random forest with some other model. Initially, I tried training Ridge, however the result wasn't that good. It was reporting a loss of around $129. I then used Catboost instead. The loss for this model was around $112, so I went ahead with that one.

download


ed-donner:
OK very nice! Thanks for the interesting experiments..

PR #898, opened 2025-10-29
Kwabena Bootcamp Week 5

kbaah:
resume expert

This solution could use some kind of eval to tell if the RAg is actually better at generating resumes but yeah it was absolute fun! And this could actually be very useful too.


ed-donner:
Yes - very nice indeed! I know something of this field!
Ed

PR #899, opened 2025-10-29
Kwabena Bootcamp Week 6

kbaah:
output

My finetuned model was actually better!


ed-donner:
Interesting - thank you Kwabena

PR #900, opened 2025-10-29
Kwabena Bootcamp Week 3

kbaah:
I am simply amazed at how hugging face makes it easy to work with these open source models! Hopefully a commercial version of this solution can help to generate some good data for training a model that can be later used to help users customize resumes.
week3-resume-jd-data-generator


ed-donner:
Great! Thanks so much Kwabena

PR #901, opened 2025-10-29
sach91 bootcamp week6 exercise

sach91:
Sachin's pull request for week 6 exercise in bootcamp.

**Try some approaches (as below) inline in the course notebook**
1. XGBoost Algorithm
2. Mix of experts (XGBoost, LinearRegression (LR), LR with Word2Vec) - Two variants: (1) Simple average (2) Average of 2 nearest predictions
3. Named Entity Recognition (NER) - Filter text to include only entities in the feature vectors
4. MinMax and Standard Scaler for normalizing word2vec embeddings

**Insights**:
1. None improved the performance from Random Forest.
2. When I applied any change over an existing approach, if not improved, the MAE deteriorated little only.
3. In traditional ML approaches, Random Forest is a strong baseline.

### DIFF
--- a/community-contributions/sach91-bootcamp/week6-exercise.ipynb
+++ b/community-contributions/sach91-bootcamp/week6-exercise.ipynb
@@ -1,0 +1,1167 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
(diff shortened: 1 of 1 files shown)


ed-donner:
Very nice Sachin, thanks for the interesting experiments..

PR #903, opened 2025-10-29
Week 8 exercise

CosmusMutuku:
**### Exercise Description:**

- Applied the agentic framework concepts to build a sentiment analysis system with modular agents for data preparation, model analysis, and LLM-based summarization.
- Used a small dataset of 5,000 customer reviews.
- Involved preprocessing text data, applying TF-IDF vectorization, and training a Logistic Regression model to classify sentiments as positive, negative, or neutral.

**### Results/Findings:**

- The initial model achieved high accuracy (86%) but showed class bias toward positive reviews.
- After balancing, results became more realistic (66% accuracy, macro F1 = 0.62), with improved recall for negative and neutral sentiments.
- Evaluation showed that longer reviews tended to be neutral or negative, while positive ones were shorter (which i believe it is consistent with real-world review patterns).

**### Screenshots:**
image
image
image
image
image
image
image
image

### DIFF
--- a/week8/community_contributions/week8_exercie (2).py
+++ b/week8/community_contributions/week8_exercie (2).py
@@ -171,14 +171,7 @@ sample_preds = clf.predict(sample_features)
 for text, pred in zip(sample_texts, sample_preds):
     print(f"\nReview: {text}\nPredicted Sentiment: {pred}")
 
-"""-->Positive reviews dominate,
-
--->The model basically learned “always say positive”,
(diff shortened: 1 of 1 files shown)


ed-donner:
That's pretty interesting Cosmus thank you - a nice twist on the dataset.
I wonder if tfidf + linear regression (or maybe a shallow neural network) would have done better at price prediction..

PR #913, opened 2025-10-30
Kwabena Bootcamp Week7

kbaah:
PEFT

I did beat the $47 with a smaller dataset; I must mention how fulfilling it is to finetune a model like this and get such great results. 

And I like the fact that Hugging face allows you to use a commit so you do not necessarily need to wait for the finetuned model to fully complete training.

 It is exciting that this beat the gpt finetuned model I built too; I know where to put my money when I have a business case.


ed-donner:
That's a really nice result, Kwabena. Congrats. I'm so happy you enjoyed it. Hey I notice this PR has about 7,000 lines of code - I think the colab saves with lots of extra info. Is it possible to export it in a way that is more size-friendly? I'll mark you complete on the sheet, but if you could replace with a lighter file that would be much appreciated.. thank you

kbaah:
thanks; done ed; let me know if this works.

ed-donner:
perfecto! Thank you

PR #916, opened 2025-10-30
Week 2 Exercise: FlightAI Assistant

philipomoigui:
A full-featured AI travel booking assistant with a Gradio UI that can check flight prices, book tickets, provide weather forecasts, convert currencies, and more. It supports both OpenAI GPT-4.1-mini and local Llama 3.2 with model switching.

image

**What I learned:**

1. Function calling with OpenAI's tools feature is incredibly powerful - the way the LLM automatically decides which tool to call based on user intent was smooth

2. SQLite integration for persistent storage was surprisingly straightforward 

3. Working with multiple API integrations (OpenWeather, ExchangeRate) taught me how to build more practical, real-world applications, but I guess it would make more sense if you own this service instead, as the AI output might depend on their own SLA

**What I enjoyed:**

1. Building something that actually feels like a real product someone could use

2. Seeing the tools work together - when a user asks about a price and then books a ticket, it all flows naturally through the conversation


ed-donner:
Terrific Philip, nice work, thank you

PR #926, opened 2025-10-31
An online banking application as an exercise in using tools with LLMs.

glafrance:
# Online Banking Assistant

This application is the Bank of Transformers online banking application.

Using this application users can:

- open an account
- add funds to an account
- withdraw funds from an account
- check the balance of accounts


ed-donner:
Really nicely done - this is the kind of contribution that elevates the entire repo. Thanks so much @glafrance

PR #929, opened 2025-10-31
Kwabena Bootcamp Week8

kbaah:
price is just right

Quite a journey; it will be interesting to add the finetuned model to the ensemble.

side note: You might need to add _twilio_ to the _requirements.txt_ file as its absence was causing some errors because i am using uv and cursor. otherwise students coud just run `uv add twilio` to fix this or comment the import out from the _deal_agent_framework.py_ file if they go with pushover to avoid the module not found error for twilio.

### DIFF
--- a/week8/community_contributions/kwabena/the_price_is_right.ipynb
+++ b/week8/community_contributions/kwabena/the_price_is_right.ipynb
@@ -1,0 +1,196 @@
+{
+ "cells": [
+  {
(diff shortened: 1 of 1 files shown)


ed-donner:
oh jeez thank you - will do. I think I removed it from the package but accidentally left it in the code somewhere. Thanks so much. Nicely done!

PR #935, opened 2025-11-01
John morrall/day4 illusion of memory

JohnMorrall:


JohnMorrall:
Adding a file to the community week1/day4. Not a project just playing with the memory illusion concept at a very basic level.

ed-donner:
oh that's very nice - such a clear & simple way for it to come together. Thanks @JohnMorrall

PR #941, opened 2025-11-02
Add carl-grp/week1 final exercise proposition

carl-graup:
Designed a very simplified tool to compare power plants for Energy systems engineering by prompting an LLM as an "energy expert" that evaluates based on costs.


ed-donner:
Nice! Great idea!

PR #943, opened 2025-11-03
Add pythonic solution for AI assistant with tools

neta-haluts:
pythonic solution to handle_tool_calls so it works with any function (without if...)


ed-donner:
Oh yeahh.. Good ol' globals()
But it perhaps would be even more pythonic to save that line of code and do func(**args) ?
Anyways you appear to have accidentally included something from outside community_contributions in this PR..

PR #947, opened 2025-11-03
Attempting to use GPT to create a Sterling Integrator business proces…

camio888:
…s using BPML. Results are hardly production-ready


ed-donner:
interesting experiment! Thanks @camio888

PR #955, opened 2025-11-05
Added week 7 solution

philipomoigui:
image

### Summary
Completed Week 7 exercise evaluating a fine-tuned LLaMA 3.1 8B model for product price prediction using QLoRA and PEFT.

### Approach
- Loaded pre-trained LLaMA 3.1 8B model
- Used 4-bit / 8-bit quantization for memory efficiency on T4 GPU
- Implemented weighted top-K prediction for robust price estimates
- Evaluated on 250 test samples


ed-donner:
Very nice - thank you so much

PR #965, opened 2025-11-11
Solved 3-chat and set the ticket price

salma-w:
I think it is correct


ed-donner:
Hey Salma!
Ahh actually it looks like this has files outside community-contributions. Please see point 1 here:
https://edwarddonner.com/pr
Many thanks
Ed

PR #988, opened 2025-11-18
CV resume checker

ohausner:
Added notebook in week1/community-contributions.

My notebook builds a model that compares a user's resume (from a pdf/word document) against a job description taken from a url online, and suggests improvements to the resume according to the job description and requirements.


ed-donner:
Aha - nice use case! Thank you!

PR #989, opened 2025-11-18
Vadi/week2 day1 advanced excercise

vadirajkrishna:
Added notebook with solution to advanced exercise where I have simulated a python interview between Claude (Interviewer) and Gemini (candidate) and then the entire interview is finally evaluated by OpenAI which gives final verdict as Gemini is Selected or Not with reasoning.


ed-donner:
Great! There's a lot in here - thanks for sharing!

PR #997, opened 2025-11-20
Kokoro Text To Speech Example

smoses2:
Kokoro is a light weight, but strong Text to Speech Library on Hugging Face [HuggingFace: hexgrad/Kokoro-82M](https://huggingface.co/hexgrad/Kokoro-82M/tree/main). 

The results are surprisingly good (good enough to not pay for a commercial product - I'm looking at you ElevenLabs).   There are dozens of different male and female voices from around the world.

Install espeak on your system, and "uv add" 3 python packages.   Instructions are at the top of the file.   

This pull request includes a short ipynb file, and its example output of 4 small wav files (<1 MB total)


ed-donner:
That's very cool - thank you so much for this

PR #1004, opened 2025-11-22
3 Way Poetry Conversation between Bing Crosby, Charlie Parker and Alice Roberts

mannnick24:
I added a conversation handler class, a character class and a conversation message class, to track the characters system prompts and the conversation system prompts and the responses from the characters 
I wrote an example usage in a notebook for a poetry competition between 3 characters Bing Crosby, Charlie Parker and Alice Roberts


ed-donner:
Great - thanks so much for sharing

PR #1018, opened 2025-11-29
fix(deps): Resolve torch installation issues on macOS Intel

rafaelmtro:
The dependency 'torch>=2.0.0' was leading to installation failures on macOS Intel (x86_64). This PR pins a known-compatible version using an environment marker in pyproject.toml to ensure successful syncing.


ed-donner:
Hey thanks for this. I think this instruction is covered in Q11 in the faq, as linked in the setup docs - or does that need to be updated? Given the number of people on the course, I have to be very careful about any changes to uv.lock or pinning versions without rigorous cross-platform testing. And then still there are often unexpected issues for about 1% of students, and that means over 1,000 emails!

raphaelmtro:
Hi! You are right. I was able to solve the macOS intel issue following the FAQ instructions. Thanks!

PR #1021, opened 2025-11-30
exercise changes

RAJESH-GN:
Hey Ed,

Here is the update on the exercise as u have asked from week 2 day 1 . Which is to have a conversation with 3 people. Please review my code and let me know if it is good.

Thanks,
Rajesh


ed-donner:
Terrific - much appreciated

PR #1031, opened 2025-12-02
3 way conversation with user instead of third LLM with some 2 seasoned philosphers

nucino:
Sorry for the troubles I caused from my last pull request, I executed some ill git commands and added already existing files, so I closed the old pull request and created this new one.

Thanks and regards!


ed-donner:
That's all good - thanks so much

PR #1040, opened 2025-12-05
Brandon day2 example

brandonlopezh:
Adding ollama modifications to day1 example. Also creating this in the right community contributions this time.


ed-donner:
Thanks much

PR #1048, opened 2025-12-07
Brandon week1

brandonlopezh:
week1 exercise! Answers question in Portuguese simply and in Shakesperean.


ed-donner:
great! Thank you!

PR #1061, opened 2025-12-12
added week 1 day 1 basketball summary

zeca77:
Just a very simple prompt that fetches some selenium extracting from other submissions and uses it to make a summary of an NBA game.


ed-donner:
Thanks so much

PR #1064, opened 2025-12-13
Community contributions - master mind game with analysis and audio

george-wiles:
As part of week 2 last exercise, I create a game (mastermind colour selection) using two llm prompts:

- one that will analyse a colour sequence guess based on past guesses and scores. 
- one that allows you to question or challenge the llm feedback analyses

the initial llm also grabs the detailed response into a text box, and the summary (last paragraph recommendation) as an audio call.

Lots of fun...would like to post in linked in and get amplified!  the analysis prompts still gets has some false positives were it incorrectly infers eliminated colours from previous guesses, so I am starting to understand the importance on getting the prompts right.. I had to keep seeding the system prompt with the game rules so the LLM does not forget...


ed-donner:
I love it. Thanks so much @george-wiles . Yes please do post on LI and tag me, and if I don't get to it in a few days then please message me with the link so I definitely get to it. Really nice

PR #1076, opened 2025-12-14
W2D1: 3-way conversation (GPT + Claude + Mistral)

guillermorosasnegron:
What this PR adds
Implements a 3-way conversation loop (GPT + Claude + Mistral)
Uses zip() with padding to avoid truncation
Uses 1 system + 1 user message per call, where the user prompt includes the full conversation so far
Files
week2/community-contributions/W2D1_3AI_conversation.ipynb
Notes
GPT via OpenAI key, Claude/Mistral via OpenRouter


ed-donner:
Great! Thank you

PR #1080, opened 2025-12-15
Add marstippo week 1 lab 5 and exercise 1 solutions (cleared outputs)

marstippo:
Added Week 1 Lab 5 and Week 1 Exercise 1 solutions with cleared outputs.


marstippo:
Cancelled as the previous file uploaded was not saved.

PR #1085, opened 2025-12-17
add day2 basketball and week1 exercise

zeca77:
Added an adaptation of simple prompt that fetches some selenium extracting from other submissions and uses it to make a summary of an NBA game, this time using llama 3.1 claude


ed-donner:
looks like you might need to clear outputs?

PR #1090, opened 2025-12-18
[Alberto Real] (W2-D5) feat: Sorting Hat LLM Experience

alberto-real:
### Contribution of Alberto Real, Related to week 2, day 5.

The goal was to demonstrate the assistant in detail, and the expected results are:

1. The user can choose which model to use (GPT or local Ollama).
2. The AI ​​simulates the famous Sorting Hat from Harry Potter. By asking three questions, it chooses one of the four houses based on the answers.
3. The process accesses a small database to retrieve the house ID based on the name of the previous step.
4. Using that ID, it makes an HTTP request to a [public API](https://wizard-world-api.herokuapp.com/swagger/index.html) to retrieve details about the house.
5. Finally, the AI ​​processes the HTTP response to match the user's answers with the response data.


ed-donner:
very fun! Thank you

PR #1092, opened 2025-12-19
Venkat tumu

vnrtumu:
**Implementation**

Added a new Day 1 Week 2 notebook showing how to use OpenRouter via the OpenAI SDK, enumerate available models, probe each model, and run a multi‑persona chat demo across three different models (Day1_week2_with_free_OpenRouter.ipynb).

Includes a simple requests-based call to list models and a loop to test each model with a sample prompt, capturing successes/errors.
Demonstrates loading OPEN_ROUTER_API_KEY from .env, setting up the OpenAI client against https://openrouter.ai/api/v1, and a multi-speaker conversation flow (Alex/Blake/Charlie) mapped to distinct models.
Testing: not executed here (no OPEN_ROUTER_API_KEY set in this environment).

Next steps: 
1) Run the notebook with a valid OpenRouter key to verify calls and update any failing model IDs. 
2) Trim or pin the model list if you want to avoid long runs or paid models.


ed-donner:
Great! Thanks

PR #1099, opened 2025-12-21
Week 2: Multimodal gemology assistant

KfirTayar:
This PR adds a notebook that implements a multimodal AI gemology assistant for diamond pricing using real-world data.

The assistant combines structured inputs and tool-based price lookup, enforces required diamond attributes before tool calls, and is served through a simple Gradio chat interface with optional text-to-speech support.

The full explanation and implementation details are provided in the notebook.


ed-donner:
Impressive project - thank you for sharing

PR #1105, opened 2025-12-23
Added week1 solutions

srj1407:
Added solutions to exercises in week1.


ed-donner:
Great - thanks so much

PR #1106, opened 2025-12-23
added Carrie's week 1 solution (cleared outputs)

cbgru:
added week 1 exercise with cleared outputs


ed-donner:
Great - thanks so much

PR #1109, opened 2025-12-24
Week1 day2 exercise

sbm4034:
Email Reply Generator

This solution takes an email and generates an appropriate reply using Llama API model.

Parameters:

    Email content (the message to reply to)
    User name (for the signature)

The code uses object-oriented design with separate methods for prompt creation, API calls, and output handling. The AI model intelligently determines the type of email (job posting, interview, etc.) and responds accordingly.

Future iterations will include improved error handling, type hints, and testing as I develop stronger Python skills.


ed-donner:
Thank you again!

PR #1112, opened 2025-12-28

srinath-19:
My project is a small end-to-end pipeline that downloads YouTube audio, transcribes it with Faster‑Whisper, generates time‑based chapters, and summarizes the transcript with OpenAI.
Thinking to also add Computer Vision in the framework this project only captures audio while the video might contain codes, diagrams, etc. I need to capture those also to make this project absolute.


ed-donner:
hey - rather than a sub-module, would you be OK to include a markdown file that describes your project, then has a link to the URL of your github repo? Many thanks

PR #1119, opened 2025-12-28
Shield01/day2 homework

Shield01:
Implemented Website summarizer using llama3.2 open source model instead of previous frontier model implementation


ed-donner:
Great! thanks

PR #1128, opened 2025-12-31
Created a commercial AI assistant chatbot

Shield01:
Created a commercial AI assistant chatbot that has the ability to retrieve ticket prices from a db and also help customers with their booking requests


ed-donner:
Thanks for all these!

PR #1138, opened 2026-01-02
Introduce instructor and social media connections

roshan-rubrik:
Added a section introducing the instructor and social media links.


roshan-rubrik:
created by mistake.

PR #1139, opened 2026-01-02
Updated system_info.py (For Windows Platform)

Akshay4452:
I observed that the older version was not able to find the _clang_ or _MSVC_ compilers available on Windows platform. It was working for Ed since he was working on his MacBook. So maybe the script works perfectly for MacBook but there were problems with Windows version. I have enhanced the method `_toolchain_block()`, it is now able to fetch clang/MSVC compilers correctly. 
**Note:**
1. This is helpful especially for people who already has Visual Studio installed (not VS Code) or if they have clang compiler installed. Need to check the working state of script for gcc/g++ compilers.
2. MacBook users may not need this update. This is particularly for Windows users.
3. Requesting peers to please check for gcc/g++ compilers.

### DIFF
--- a/week4/system_info.py
+++ b/week4/system_info.py
@@ -28,6 +28,7 @@ def _first_line(s: str) -> str:
 
 
 def _which(name: str) -> str:
+    # shutil.which returns the absolute path if 'name' is already a valid absolute path
     return shutil.which(name) or ""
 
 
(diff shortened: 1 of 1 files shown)


Akshay4452:
Requesting Ed to review the code changes and update the python script if my changes are helpful.
Thanks and regards,
Akshay Manthekar

ed-donner:
Thanks Akshay.. I'll probably take some more time on this to regression test it on a Mac first.. thanks so much

PR #1144, opened 2026-01-02
Added my contributions to community-contributions for week1 day1 lab

jebeckford:
I created a notebook that uses playwright to scrape website content and I had to introduce some exception handing and conditional logic, in addition to cookies to be able to bypass bot blockers.  The idea was to scrape a website that lists businesses (bizbuysell) and create a list that includes the names of the businesses, pricing, revenue, cash flow and the best approach to validate the value of that business.  This is pretty raw and I plan to create an agent that runs this regularly for me but it's a start.


ed-donner:
Thanks so much

PR #1149, opened 2026-01-03
Created a synthetic data generator

Shield01:
Created a synthetic data generator, that stores the data generated to a file


ed-donner:
Great! Thanks so much

PR #1156, opened 2026-01-04
added ragu_notes.md file as test as a part of day1 class

ragu-selva:


ed-donner:
the test looks good - are you ok to close this?

PR #1163, opened 2026-01-05
Added 3 way conversation

sbm4034:
This NB file showcases a three way fun virtual conversation between Messi, Ronaldo and Mbappe.


ed-donner:
Great!

PR #1164, opened 2026-01-05
Add community contribution RegressionHeadPricer

antonawinkler:
I have been experimenting with pricers which derive prices from embeddings. This notebook demonstrates my best results so far. I would like to share it with the community.


ed-donner:
Oh very nice indeed - thank you for sharing

PR #1174, opened 2026-01-07
community - day1 - playwright + categorize websites

abdennour:
instead of using OpenAI to summarize website homepage, this exercise will categorize website: 

- Search Engine
- eCommerce
- Sports
- Software Dev.
- ...etc

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -213,3 +213,5 @@ week6/*/*.jsonl
 week6/*.pth
 week6/*.joblib
 week8/*.pth
+# ignore node_modules
+node_modules/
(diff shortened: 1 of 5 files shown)


ed-donner:
hey this seems to make an update to the repo .gitignore?

abdennour:
yes @ed-donner  i added "node_modules" in .gitignore which is standard line in any repo has nodejs code.

ed-donner:
Hey @abdennour,
Even in the case of node_modules, the standard is that if your project has nodejs code, then you would add a .gitignore to your subdirectory. I've merged about 200 PRs today; yours is not the only to add something to .gitignore (and to other similar repo level files like uv.lock). Imagine the change management if the repo-wide .gitignore changed every day; and one day, one of them would have an unintended side-effect.
With many thanks
Ed

abdennour:
@ed-donner yes that make sense ! I will do

PR #1181, opened 2026-01-08
Meeting Summarizer - MoM

swbhalla:
Program to summarize a meeting transcript and create minutes of meeting


ed-donner:
Great - thank you

PR #1182, opened 2026-01-08
Add Madhura's lab1 solution

madhurashinde:
1. Takes content of an email and provides subject line
2. Generate data science related instagram post from website content


ed-donner:
Very nice & useful! Thank you

PR #1185, opened 2026-01-09
create a music booklet for kids based on input music sheet png files

Cecilina:


ed-donner:
That's an interesting, novel project! Thank you

PR #1189, opened 2026-01-11
top buzz words used on eds website (ollama way)

yotalha:


ed-donner:
Another nice one - thank you

PR #1200, opened 2026-01-15
[Alberto Real] (W4-Ex) feat: Job offer and CV skills comparation

alberto-real:
A Gradio view for easily uploading a CV and a job offer.
A dropdown for a selection of models: frontier or local.
An execution button to view a summary of key skills, years of experience, and a list of skills not found.


ed-donner:
Thanks! Nice example..

PR #1201, opened 2026-01-15
added community code to fetch a story and convert to english from japanese

abhi1992:
Fetch a story in japanese text and convert it to english.


ed-donner:
nice! short & useful. Thank you!

PR #1214, opened 2026-01-17
added litellm in requirements.txt

rohit141914:
I was getting  error in "from litellm import completion" then i added "litellm" in requirements and run pip install -r requirements.txt then it works fine

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -36,3 +36,4 @@ sentence_transformers
 feedparser
 protobuf==3.20.2
 wandb
+litellm


ed-donner:
Hello Rohit,
No; it looks like you may have misunderstood.
In this course, we use a powerful package manager called "uv", which is very popular in the industry.
The kernel for notebooks uses the uv environment.
LiteLLM is included in the uv project - see pyproject.toml.
With uv, never "pip install", always "uv add".
In the course, all packages are provided for you in the uv environment. If a package is missing, that's your clue that something is wrong with your environment.
If in doubt, please review the Setup guide, and in particular the instructions on uv.
Many thanks
Ed

PR #1215, opened 2026-01-17
Add Francisco's week2 day5 contribution

fberruet8:
In llm_Engineering week2 day5's project I realized that the assistant would always render an image of the second city provided on the compare two city prices query instead of rendering the image of the cheaper one.

I though that perhaps this might be solved by using tools to rendering the images rather than a plain function call in the chat method.

Since it sounded like a good tools exercise, I decided to jump into it and created the modified code below.

It works fine from a tools perspective but unfortunatelly when I ask the agent about a cheaper option between two cities it does not render any image at all even though instructed to do so in the system prompt.

Could you please provide some insight about what might be going on and suggestions about how to better promt it?

Thanks


ed-donner:
Hey! Do you want me to merge this anyway?
If you want to work with, please see my comments in Q35 here:
https://edwarddonner.com/faq/
Many thanks
Ed

fberruet8:
Hi Ed - thank you for the response. No, please do not merge it - Let me work a bit on the system prompt and I will re-submit

Thanks

PR #1252, opened 2026-01-23
first commit to the course! 

ColinPanarra:


ed-donner:
Hmm I'm not sure what you're doing here.. this is a markdown file rather than a jupyter notebook, but with a .ipynb extension?

PR #1275, opened 2026-01-27
Added my code to summarize a website using playwright and anthropic

sneharao:
The scraper now uses Playwright's Async API, which is compatible with Jupyter's asyncio event loop.
Instead of OpenAI have used Anthropic API


ed-donner:
OK thanks v much

PR #1277, opened 2026-01-27
Week2 day5 Francisco's contribution - Image creation tool

fberruet8:
### Week2 day5 Francisco's contribution: 

Since day 5 code always create an image of the first city in a prompt, when asked about a price comparation (cheaper or more expensive) it always render an image of the first city in the prompt. To make sure that the image rendered corresponds to the most relevant city, I modified the code to implement the image creation via a tool.

Also changed the model from gpt-4o-mini to gpt-5-mini because, although I tried multiple prompts, I did not manage to have gpt-4o call multiple tools in a sequenced order - which is the case when having to first figure out the prices difference and then request the image of the most relevant one.


ed-donner:
Interesting - thank you

PR #1285, opened 2026-01-28
llm debate 3way solution for llm chat

Bichesq:
Deepseek and Sarvam on a debate standoff - Sarvam is a free Indian LLM model I found and so using for this course along with deepseek. Deepseek also doubles as the moderator.
Purpose: To demonstrate debate orchestration with multiple LLMs in the course.


ed-donner:
Great - thanks much

PR #1291, opened 2026-01-31
[EXERCISE] three way chat with final fantasy characters

jgandrade:
Exercise for Three way LLM conversations simulating Final Fantasy character interactions using DeepSeek R1. Features FF7 and FF9 character trios with personality-driven dialogue generation.


ed-donner:
Nice! Thank you!

PR #1296, opened 2026-02-01
elasticsearch-explorer

mannnick24:
Elasticsearch explorer
The aim is to input your unauthenticated elasticsearch url and this tool will help you query it with natural language.
You can ask "please aggregate all agent status records by companyId for the last 6 months" and you will get the results in a table - saves me looking up the syntax all the time
there are 2 tools for reading the searchable aliases and finding the mapping for each. You can ask to "list all aliases" and you can ask to see the mapping for a particular alias. you can also just ask for the field list for an alias.
If you ask, you can just get the search query displayed - usefull for debugging
If you ask to query for records, the llm will use the tools to first create a query, then execute it and show you the results formatted to a table.
The date and the "companyId" if you are using it will be set dynamically into the prompt


ed-donner:
Oh wow this is super useful - nicely done!

PR #1299, opened 2026-02-02
add lavi week 1 exercise

lavigarg05:
Added a technical assisstant for selenium which answers selenium based technical questions for users.
It uses openai and ollama based on user preference.


ed-donner:
nice!

PR #1302, opened 2026-02-03
my community contribution - an alt text generator using computer vision

shaxbeard:
I have extended the week 2 content by making a call to computer vision APIs instead of chat.

I have used the Gradio dropdown component to select between OpenAI and Anthropic providers from the same UI.


ed-donner:
That is handy! Thank you

PR #1306, opened 2026-02-03
Added summarization using ollama

sneharao:
Added summarization using model llama3.2


ed-donner:
hey - is this editing your own project, or somebody else's? I can't see..

PR #1316, opened 2026-02-05
Fixed the HAZARDOUS COMBO of while loop and list's extend method. 

mahendra0120:
Also added the streaming ability to FlightAI Assistant


ed-donner:
Nice! Thank you!

PR #1317, opened 2026-02-05
[Week 2 Community Contribution] Draw my story notebook source code

ZombieDadCoding:
# Project - Draw my story

Create an interactive story with a few prompts from a toddler and create a fun story, with interactive prompts that are stored using tools and with a bonus super fun image at the end!


ed-donner:
oh that's a super fun idea. Thank you!

PR #1333, opened 2026-02-09
Added a feature for website comparison on the summarizer of day 1 project

AliDavoodi98:
In this PR I have tweaked the summarize function to have the option to obtain two different urls and would be instructed to create a table of comparison between citing their primary focus, feature type, and more.


ed-donner:
Great - many thanks

PR #1339, opened 2026-02-10
Added Sourav Sinha  Week1 Exercise solution (cleared outputs)

Phoenixdelta74:
End of Week 1 Exercise has been added Built a Tool based on the constants .


ed-donner:
Many thanks

PR #1340, opened 2026-02-10
rag week5 assignment obsidian vault Q&A text agent

khudgins:
Scans an obsidian vault - in my case a D&D campaign - and provides a query agent to help run a game.


ed-donner:
very cool! Thanks!

PR #1341, opened 2026-02-10
DEPRESSED MAN VS OVERLY ENTHUSIASTIC WOMAN VS SMUG DUDE

mahendra0120:
Here's my slightly new take on 3way conversation between three LLMs.
Hopefully you wont find any problems with this.


ed-donner:
Thank you!

PR #2952, opened 2026-03-31
Week2 day4 solution

ashutoshthakur454:
I have added the tool called set_ticket_price. Alongside that I have added some print statements to develop my understanding of the logic flow between the llms and the tools.


ed-donner:
Great - thanks so much!

PR #2970, opened 2026-04-02
Adding quant vs non-quant Llama response on T4GPU

Tuckker:
This project explores the impact of quantization on Large Language Model (LLM) inference by comparing the response quality and behaviour of a quantized versus a non-quantized model loaded locally using Hugging Face Transformers. The experiment loads a LLaMA-based model in two configurations — one using 4-bit quantization via BitsAndBytes and one in full precision — and runs both against the same input messages. Tokenization is handled manually using AutoTokenizer with a chat template applied to format the input into the model's expected prompt structure before being passed to the GPU. The outputs are decoded back to human-readable text and compared side by side to evaluate any differences in response coherence, accuracy, and verbosity. The goal is to understand the trade-offs between memory efficiency gained through quantization and the potential impact on model output quality, providing practical insight into how quantization can be leveraged for deployment in resource-constrained environments such as Google Colab with a T4 GPU.


ed-donner:
Hey,
I'm super grateful that you've worked on this and are sharing your contribution. It looks like this might not be ready to be merged - please can I ask you to review this:

https://edwarddonner.com/pr

The most common issue is that you've included a file outside community-contributions. This would impact all students on the course. You can see this by going to the "Files changed" tab in this PR; the only changes should be to files in community-contributions.

If there's a good reason to change files outside community-contributions, like an important bug fix, then please let me know (and thank you!)

The 2nd most common reason is that you've not cleared outputs of Notebooks, leading to large code changes. Instructions at the link above.

The 3rd common reason is that you've included too many changes, such as many thousands of lines of code, or a large number of files. I'm grateful for your work, but see the link above for ways we can include this without the repo getting too large.

The PR details in this Github screen shows you the files changed, and at the top right the number of lines added (in Green) and deleted (in Red) by this PR. Typically the Green number should be less than 2,000 and the red number should be 0.

I very much look forward to merging when this is addressed. Thanks so much
Ed

Tuckker:
Hi Ed, 

Thank you for the feedback and help , I have placed the file in the correct folder which is community contributions now.

Many thanks

ed-donner:
there are still changes in the PR outside community-contributions

Tuckker:
Hi Ed, alright now the copy file has been removed from outside community contributions.
Many thanks

ed-donner:
Great - thanks so much!

PR #2976, opened 2026-04-03
week 2 day 4 exercise

shahanabahlul:
Added a solution for exercise on week2-day4. Updating the DB via UI is not ideal, but I wanted to try it.


ed-donner:
Great - thanks so much!

PR #2987, opened 2026-04-04
Add week1 day2 solution (community contribution)

Adityaz23:


ed-donner:
Great - thanks so much!

PR #2988, opened 2026-04-04
(kacper-lechicki): week2 -> model arena

Kacper-Lechicki:


ed-donner:
Great - thanks so much!

PR #2992, opened 2026-04-05
Add benchmark module with streaming metrics, auto timing, and stale file detection for week 4 day 3

Himesh-29:
@ed-donner 

Replaced the scattered port/compile/average cells in day3 with a single benchmark.py module for week 4 day 3 jupyter notebook

- Auto-measures Python baseline time instead of hardcoding it
- Uses streaming to track TTFT and total generation time per model
- Hashes main.cpp before/after each model to catch stale responses
- Raises on API errors instead of silently reusing old main.cpp
- Prints a ranked comparison table with speedups at the end

image
image


ed-donner:
Great - thanks so much!

PR #2993, opened 2026-04-05
explains the mistake in the generated code and fixes based on that

shahanabahlul:
Built upon the week 4 day 5 exercise, when the generated code is incorrect, LLM gets the error message and explains it in the ui, then, if needed, regenerates based on that explanation and runs the new code.


ed-donner:
Great - thanks so much!

PR #2994, opened 2026-04-06
Implemented three LLMs conversation

sebahadin:
Implements a multi-LLM conversation with distinct roles and real-time streaming responses.


ed-donner:
Great - thanks so much!

PR #2995, opened 2026-04-06
week two exercises

erybie222:


ed-donner:
Great - thanks so much!

PR #2996, opened 2026-04-06
3-way LLM Conversation Exercise

michaeldimimu:
- A conversation between 3 popular psychologists with differing views on the motivation behind human behaviour (Alfred Adler, Carl Jung and Sigmund Freud)
- Used OpenAI's GPT 4.1 Mini as the model for all characters


ed-donner:
Great - thanks so much!

PR #3001, opened 2026-04-07
Find the theme of a website

bezaleluriel:
Added a notebook that uses Ollama to detect a website theme.


ed-donner:
Great - thanks so much!

PR #3003, opened 2026-04-07
STEM teaching

springplat:
When it comes to teaching young students coding and computational thinking skills project based and personalized learning are key. I'm excited to learn and leverage AI in this area and see how i can meet every student where they are.


ed-donner:
Great - thanks so much!

PR #3010, opened 2026-04-08
Week2 - Meal Planner Assistant with Tool calling

gitkacha:
1. Meal Planner using recently released Gemma4
2. Uses a sqlite db to store the data in two relational tables - Recipes and Meals
3. When the user confirms, it first saves recipes and then saves weekly meal plans referencing those recipes in a two step process


ed-donner:
Great - thanks so much!

PR #3019, opened 2026-04-11
Add an Article Maker in Week1

sanskritiigarg:
An Article Maker for content creators, which takes the title and URL, transcribes the video and creates an article that the creator can post on their blog. This uses Gemini 3.1 Flash Lite.


ed-donner:
Great - thanks so much!

PR #3025, opened 2026-04-12
Renzo's week1 solution: Fundamental Analysis Expert

rca-1p:
Created a new jupyter notebook with all the necessary code to set up the system and user prompts to set up a fudamental analysis expert agent that will predict the value of a stock by the end of the year.
The response is formatted as Markdown for clean displays.

Doesn't need any new libraries other than the ones already set up with uv.


ed-donner:
Great - thanks so much!

PR #3057, opened 2026-04-17
Add initial implementation of the Hospital-Ashir RAG system

ashir1234:
Screenshot 2026-04-18 at 1 17 22 AM


ashir1234:
@ed-donner  this is the highest score i was able to achieve after modifications after the week 5 - Day 4 excercise

ed-donner:
Impressive! But please could you remove the knowledgebase / md files from the PR; instead, provide a link to download them from somewhere central like google drive.. otherwise the repo will grow to an unmanageable size.. many thanks

PR #3058, opened 2026-04-18
Add week1 exercise

swattingmoth:
Add week1 exercise. 
Use an LLM to determine the best model to answer a question, then use that model to answer to the question.


ed-donner:
Great - thanks so much!

PR #3065, opened 2026-04-19
created a conversation agent flow with three different models:

himanshukr02122:
1. openai - gpt-5-nano
2. gemini - gemini-3-flash-preview
3. ollama - deepseek-r1:1.5b

all the above agents are communication in random order.


ed-donner:
Great - thanks so much!

PR #3066, opened 2026-04-19
Feature: Week 1 Final Exercise - User Input for 'question' Variable

varshavijay02:
# Summary 
Instead of having 'question' as a variable with a constant value, I have made changes to get input from the user.

# Changes
Removed the constant assignment of question and added logic to accept input from the user at runtime (e.g., via CLI/input prompt)


ed-donner:
Great - thanks so much!

varshavijay02:
Thank you Ed for the wonderful course :grin:

PR #3070, opened 2026-04-21
scraping and connecting result with openAI api

ronnyp07:


ed-donner:
Great - thanks so much!

PR #3072, opened 2026-04-21
Add community contribution: Technical weblink finder and ranker

hiran1234:
Two-step GPT-4o-mini pipeline that finds weblinks for a technical topic, scores each for relevance using structured JSON output, and surfaces the top results with a topic explanation.


ed-donner:
Great - thanks so much!

PR #3073, opened 2026-04-21
Week 7 Start Capstone Project: The Price is Right

spasquier:
My fine tunned model is available at: https://huggingface.co/salvajpa/price-2026-04-21_21.32.03-lite/commit/3323e825b5c1cce08e8c6305cd17fdef605dec4d

It performs at least better than a human, and only $1.45 more than gpt-4.1-nano.

**Also is better than the $65.40 result from the Fine-tuned Lite model in the original lab.**

Screenshot 2026-04-21 161737


ed-donner:
Great - thanks so much!

PR #3074, opened 2026-04-21
Week 5 Exercise: Ruby on Rails expert with RAG

spasquier:
This submission is a Ruby on Rails expert that uses the official documentation from versions `7.2` and `8.1`.

It implements an advanced RAG solution that has smart chunking using a frontier model, rewriting the user query to get better results and re-ranking the retrieved context to generate better answers.

Screenshot 2026-04-21 170213


ed-donner:
Great - thanks so much!

PR #3077, opened 2026-04-22
Created a 3-way conversational system between multiple LLMs, where three different AI agents interact in a shared football dialogue.

tonkataaa:
Each model represents a distinct persona:
- **GPT** – argumentative Real Madrid fan  
- **Ollama (gpt-oss)** – polite Arsenal fan  
- **Qwen** – passionate Liverpool fan  

## How it works
Each model:
1. Receives the full conversation history  
2. Sees the other agents as external participants (user messages)  
3. Responds according to its own personality  

The conversation is executed in a loop

What do you think about it, do you like it?


ed-donner:
Great - thanks so much!

PR #3081, opened 2026-04-22
week2 solutions (cleared outputs)

aswamina:
week2 solutions after clearing outputs


ed-donner:
Great - thanks so much!

PR #3082, opened 2026-04-22
suveerchaudhary/week1 solution

suveerchaudhary:
My own edits/test with week1 files.
Solution - week1_EXERCISE.ipynb.


ed-donner:
Great - thanks so much!

PR #3088, opened 2026-04-23
Commit after week 2 final exercise

Samar3007:
Implemented a news assistant that retrieves and presents real-time headlines using a news API from "https://newsapi.org"


ed-donner:
Great - thanks so much!

PR #3097, opened 2026-04-26
Adding new day 2 solution inside of new subfolder

nate-lev:
Adding new day 2 solution inside of new subfolder


ed-donner:
Great - thanks so much!

PR #3101, opened 2026-04-27
Add political debate chatbot with 3 bots (Steve, Pete, Dean)

aerahn:


aerahn:
Debate chatbot with 3 participants

ed-donner:
Great - thanks so much!

PR #3104, opened 2026-04-27
Add RAG vs Finetuning files

mukherjee-ayan:


ed-donner:
Hey! Thanks for this.. would it be possible to exclude binary files from the PR, and see if it can be trimmed down anywhere? Thank you..

mukherjee-ayan:
@ed-donner I have removed the binary files from the PR. Thanks for pointing out.

ed-donner:
Great - thanks so much!

PR #3106, opened 2026-04-27
Week 8 - Example Agentic Workflow with Code Generation Agents

spasquier:
# Week 8 Exercise

## Agentic Workflow with Code Generation Agents

This notebook shows an example of two Agents (Senior Architect and Senior Developer) working in sync to ship a feature in ZIP format containing the code that fulfills the user request.

You can test it with any high level query and it will create a Ruby on Rails app that fullfills it.

**Example query to test it: Create a TODO list app.**

_Another more difficult query: Create an online Shop with Products, Categories, Orders and Customers including a REST API._


ed-donner:
Great - thanks so much!

PR #3108, opened 2026-04-28
feat: oluwadara - week 1 exercise

OluwadaraDaily:
## What is in this PR?
I worked on the week 1 exercise which is to create a simple tool to explain a technical solution


ed-donner:
Great - thanks so much!

PR #3109, opened 2026-04-28
add multi-model agent with tool calling for community contributions

jsjasee:
Added week 2 day 5 assignment which is a comprehensive multi-modal AI language tutor specialising in chinese, with the ability to remember and test words that user has learnt before.


ed-donner:
Great - thanks so much!

PR #3118, opened 2026-04-28
Week2 day2 solution

Adityaz23:
## Can build the brochure by entering the URL of the website
**Used two models one ollama local model and another gemini model**


ed-donner:
Great - thanks so much!

PR #3125, opened 2026-04-29
Danilo lab1 solution

danilojpl:
A simple llm call to describe the planets in the solar system.


ed-donner:
Great - thanks so much!

PR #3131, opened 2026-04-30
Community project summary

danedyy:
week 1 task:
Scrape community project dir
Picked random 3 dir (to reduce tokens),
parse to string and give llm to summarise projects
return result in a markdown table with author name, path and summary


ed-donner:
Great - thanks so much!

PR #3139, opened 2026-05-02
Added week3's excercies solution

ashutoshthakur454:
This notebook uses two llm calls method to create a conversational chatbot specialized in creating datasets for different sorts of supervised (Classification & regression) and unsupervised (Clustering) task. There is a tool present which creates datasets. This tool uses an llm to output a specific json format which is then used to create the dataset. Experiments were done on model selection where gpt 4.1 mini frontier model was compared with various open source models : LLAMA, QWEN, PHI, GEMMA and DEEPSEEK. For the brain llm, which would interact with the user, ask clarifying questions and give suitable modified prompt to the tool, gpt model was used as its much more powerful than the open source models.


ashutoshthakur454:
I have cleared the cell output then too github says 12000 lines changed. I am not very sure why so many lines have been affected. Is it because I have loaded multiple opensource models from hf and downloaded the notebook from colab? Keen to know your opinion

ed-donner:
Hey! Thank you for sharing this. I believe there's a way to export colabs from google that removes outputs and metadata.. this one has 12,000 lines! Would that be possible? Thanks so much

ashutoshthakur454:
Hey Ed, I have cleared the cell output of my colab notebook, re downloaded and re uploaded it. The PR now has 800 lines instead of 12000.

ed-donner:
Great - thanks so much!

PR #3142, opened 2026-05-02
add visualizers for different models to compare tokenization

aswamina:
tried to visualize how different models tokenize .   but found out that anthropic and groq do not support log_probs.  I want to move on with the course but visit this later to explore other ways of visualization.


ed-donner:
Great - thanks so much!

PR #3161, opened 2026-05-04
Week 2 Exercise Soltion: AI Flight Assistant

antonKenderov:
# Created Flight Assistant with the following abilities:

### - can make a booking which is added to the database.


ed-donner:
Great - thanks so much!

PR #3164, opened 2026-05-04
Ollama summarizer

franzonia137:
Upgraded summarizer to use Ollama locally instead of OpenAI.

Features:
- Local LLM (llama3.2)
- No API usage
- Same summarization pipeline


ed-donner:
Great - thanks so much!

PR #3166, opened 2026-05-06
Add Arpit's first day exercise solution

imarp8:


ed-donner:
Great - thanks so much!

PR #3174, opened 2026-05-07
Week 2

ShivamH1:
Week 2 work:

1. AI Chat System for an airline using the OpenRouter API
2. Exercises  (assignments)


ed-donner:
Hmm this is 6,000 loc and contains updates to every lab.. was that your intention? It's more common for people to submit the end of week exercise and 1 or 2 along the way. let me know..

ShivamH1:
Yes, you're right; it wasn't necessary.

PR #3180, opened 2026-05-08
Week 4 lab solutions (cleared outputs). Fixed errors with the C++ cod…

aswamina:
…e generation and Rust code generation, by modifying the system prompts. In case of the C++ code it was to not generate the #include <bits/stdc++.h> and for Rust to replace :.6f with :.6


aswamina (commented):
moved files from main community-contributions folder to the one under the week4 directory

ed-donner:
Hey this is really great.. to keep the size manageable, please could you submit your files either in the week4 community-contributions folder, or in the top level one, but not both? Thanks so much

aswamina:
yeah thanks. I removed the one from the top level community-contributions folder. That folder is confusing btw since there is one specific to each week

ed-donner:
Great - thanks so much!

PR #3181, opened 2026-05-08
Add week2 exercise technical Q&A chatbot with Gradio, streaming, and tool use

Krispy25:


ed-donner:
Great - thanks so much!

PR #3186, opened 2026-05-09
Create Week1-FootballMatchSynopsis.ipynb

farrasezra05:
Exercise project of Week 1 Lesson - basics of multiple LLM calls
In this project, I created a pre-match synopsis generator that extracts from the two clubs' webpages (I approached this using transfermarkt.com page), whereby there are two LLM calls involved:
1. To filter the relevant links, and
2. To create the match synopsis


ed-donner:
Great - thanks so much!

PR #3192, opened 2026-05-11
Created a helpless assistant

alessandrofugazza:
An AI chatbot that doesn't have a clue (see system prompt)

Spent the better part of an hour laughing at its responses


ed-donner:
Great - thanks so much!

PR #3204, opened 2026-05-13
modified week5 day2 and day3 solutions to use supabase instead of Chr…

aswamina:
…oma (cleared outputs)


ed-donner:
Great - thanks so much!

PR #3211, opened 2026-05-15
Add files via upload

kubamjarecki:
I though i will be clever for creating another API LLM call, for AI to check about halucinations. In day 5 exercise, as You mentioned, when it gets witty, it starts to imagine things. So easiest idea was, to ask AI, to counter check if everything is in check.  Well it didn't go well. LLM didn;t find any discripencies.  But this idea has been going through my head for a longer time. As halucination is a big problem I remember when I spend whole day trroubleshooting new playbook, because I used option, that GPT told me, is for sure solid, I thought, that maybe  let AI check each other, and by many  iterations get, an answer which is bullet proof

### DIFF
--- a/community-contributions/Dugsnaga/day5.ipynb
+++ b/community-contributions/Dugsnaga/day5.ipynb
@@ -1,0 +1,517 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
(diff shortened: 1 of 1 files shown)


ed-donner:
Great - thanks so much!

PR #3217, opened 2026-05-17
Add Horus Heresy x Dressrosa community contribution

prajithmalepati:
A day 1 system prompt experiment where the model plays Ezekyle Abaddon of the Luna Wolves as first mate, with an Imperium warband washed up on Dressrosa during Doflamingo's reign. Fun lore-collision demo for flavor stacking in system prompts.


ed-donner:
Wow that sounds creative!
Great - thanks so much!

PR #3226, opened 2026-05-19
Add tiktoken to requirements

Ramin2c:
Required first time by week1/Day4


ed-donner:
That sounds like you've gone off track. This is a uv project, per the setup instructions. A uv project doesn't use requirements.txt. The file is there for backwards compatibility with a prior version of the course.

Ramin2c:
True, thanks for the answer. I undestand now after playing a bit with UV environment.

PR #3231, opened 2026-05-21
Add week3 project from day5 product with Gradio

Krispy25:


ed-donner:
There are ways to export a colab with outputs and widgets cleared so that it is under 1,000 lines of code - this is over 8,000 lines..
Thanks much
Ed

PR #3239, opened 2026-05-23
Rafay/day2 solution

rafay-840:


ed-donner:
Great - thanks so much!

PR #3243, opened 2026-05-24
Cg week2 project

CGodl:
Added my work for the week 2 exercise covering upgrading the week 1 project to include audio


ed-donner:
Great - thanks so much!

PR #3247, opened 2026-05-24
A single python file that does not depend on scraper to analyse the we…

jazgitt:
# What is different ?
A single python file that does not depend on scraper.py provided by Ed. Use this to analyse the website and gets info directly from local LLM. This is not a notebook but a simple .py file that engineers can run in their VS code directly. 

# Setup
Add your local LLM URL in .env file OLLAMA_BASE_URL=
Play with system prompt and user prompt and model name as per your convenience
Change the website as you want to experiment


ed-donner:
Great - thanks so much!

PR #3249, opened 2026-05-24
This is the tool calling for booking the train tickets with the gradio chat interface

Adityaz23:


ed-donner:
Great - thanks so much!

PR #3255, opened 2026-05-25
PDF summarizer using ollama local LLM (Gemma3)

hmnhGeek:
A generic PDF summarizer using `gemma3:270m` and `gemma3:1b` models.


ed-donner:
Great - thanks so much!

PR #3265, opened 2026-05-27
add week2 day2 brochure solution in gradio

Mon711:


ed-donner:
Great - thanks so much!

PR #3271, opened 2026-05-28
Implemented multi rewritten queried RAG

ashutoshthakur454:
Updated the fetched context function as following:

```
def fetch_context(original_question):
    new_chunks = []
    for i in range(3):
        rewritten_question = rewrite_query(original_question)
        new_chunks.extend(fetch_context_unranked(rewritten_question))
    og_chunks = fetch_context_unranked(original_question)
    chunks = merge_chunks(og_chunks, new_chunks)
    reranked = rerank(original_question, chunks)
    return reranked[:FINAL_K]


```
    
With this change, 3 rewritten queries would be formed and 3x10 sets of chunks would be merged with the original 10 chunks and finally after reranking, 10 most relevant chunks would be chosen

Below is the eval results corresponding to this change: 
image

### DIFF
--- a/community-contributions/ashutosh/Week5/Day5/answer.py
+++ b/community-contributions/ashutosh/Week5/Day5/answer.py
@@ -1,0 +1,147 @@
+from openai import OpenAI
+from dotenv import load_dotenv
+from chromadb import PersistentClient
+from litellm import completion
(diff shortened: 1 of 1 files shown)


ed-donner:
Great - thanks so much!

PR #3281, opened 2026-05-29
Add Brian Vogon Poetry (cleared outputs)

BrianSmithPhotos:
Example web summarizer that gives output in HitchHikers style Vogon Poetry - the 3rd worst poets in the Universe.


ed-donner:
Hahaha
Great - thanks so much!

PR #3282, opened 2026-05-29
Python to Julia Code Conversion and examples - pi estimation

BrianSmithPhotos:
Code conversion to Julia, along with samples from the different models.  Julia can be added to Cursor and kernel changed to run cells, or the jl files executed at a terminal if Julia is installed. https://julialang.org/


ed-donner:
Nice! Thank you
Great - thanks so much!

PR #3284, opened 2026-05-30
Add ceema's day 2 exercise using ollama in openaI chat completeion API

Ceemamatters:
My second day exercise using Ollama api endpoint similar to openai chat completions endpoint
ollama.chat.completions.create(model="llama3.2:1b", messages=messages)


ed-donner:
Great - thanks so much!

PR #3287, opened 2026-05-31
Add Week 1 exercise: first OpenAI API call from scratch

vincjcheckmate:
Week 1 exercise building the core OpenAI chat completion pattern from scratch:
system prompt, user prompt, messages list, API call, and extracted response.
The model explains why API keys should never be committed to GitHub.
Stored under vincjcheckmate/ folder.


ed-donner:
Great - thanks so much!

PR #3292, opened 2026-05-31
Modify ingest.py

Bleufin:
## Summary

This PR adds a community contribution showing a safer version of the ingestion script under pro_implementation folder that creates OpenAI embeddings in batches.

The original version sends all chunk texts in a single embeddings request, which can fail when the input array contains more than 2048 items. The modified version updates `create_embeddings()` to process chunks in smaller batches before adding them to the Chroma collection.

## Notes

* No original course files were modified.
* The default embedding batch size is set to 500 to stay safely below the OpenAI API input limit.
* The modified file also updates the LiteLLM model from openai/gpt-4.1-nano to openai/gpt-5.4-nano; however, the main purpose of this PR is the batched embeddings fix.


ed-donner:
Great - thanks so much!

PR #3293, opened 2026-05-31
Week 1 day 1 exercise solution

diogo-alves-21:


ed-donner:
Great - thanks so much!

PR #3297, opened 2026-06-01
update price_is_right code to truly run every 5 minutes

jsjasee:
The original price_is_right.py does not run every 5 minutes because Gradio's timer function does not work reliably on my machine. Hence I ported the code over to use python's threading package instead, so that the 5 minute loop runs on the python thread on my machine, which is more reliable, instead of depending on Gradio for this functionality. Now the Gradio thread just updates the UI every 5 seconds, and is not involved in the agent loop


ed-donner:
Great - thanks so much!

PR #3304, opened 2026-06-01
homework assignment day 2

AshuT930:
# HOMEWORK EXERCISE ASSIGNMENT

Upgrade the day 1 project to summarize a webpage to use an Open Source model running locally via Ollama rather than OpenAI


ed-donner:
Great - thanks so much!

PR #3309, opened 2026-06-02
Week 1 exercise submission from Himanshu Sharma

hmnhGeek:
Solution to week 1 challenge.


ed-donner:
Great - thanks so much!

PR #3310, opened 2026-06-02
My version of a stateful chatbot

alessandrofugazza:
Uses context to simulate memory.


ed-donner:
Great - thanks so much!

PR #3315, opened 2026-06-03
Added the pdf analyser code in the day1 notebook

farru46:
Added the code snippet for analysing pf documents


ed-donner:
Great - thanks so much!

PR #3322, opened 2026-06-04
Add divide-by-zero explainer using "gpt-4.1-mini"

CodeMaestroMS:
Implements a simple GPT API intergration that explains why dividing by zero is impossible using intuitive real-world analogies.

- Uses gpt-4.1-mini via the OPENAI SDK
- System prompt instructs gpt to explain math through stories, no jargon
- User prompt asks for an intuitive explanation of division by zero


ed-donner:
Great - thanks so much!

PR #3324, opened 2026-06-04
Local chat bot that uses local models from ollama.

MK-Manikandan:
ReadMe.md is provided with details.


ed-donner:
Great - thanks so much!

PR #3325, opened 2026-06-04
🐛  Fixes a bug where Groq returns an empty response in the UI 

acjones27:
Because Groq technically uses a model with "gpt" in the name but with no reasoning effort parameter, it shows an empty response in Gradio (although it still charges you and shows up in the billing dashboard on Groq) 

This fixes it by just defining the models that can have high reasoning effort, rather than using a fuzzy match for "gpt".

Hope this helps someone!

### DIFF
--- a/week4/day4.ipynb
+++ b/week4/day4.ipynb
@@ -262,7 +262,8 @@
    "source": [
     "def port(model, python):\n",
     "    client = clients[model]\n",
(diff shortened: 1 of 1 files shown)


ed-donner:
oh jeez silly me.. thank you. There are rather better ways of doing this without us hardcoding janky strings, but this will definitely do for now! Thanks again

PR #3337, opened 2026-06-06
A simple ollama chat application created using streamlit

deepak904:


ed-donner:
Hey @deepak904 - so it's generally a red flag to me when the amount of comments and markdown is greater than the code. It's good to use Coding Agents to help, but it's a distraction if it has 100+ lines of documentation (and irrelevant instructions like virtualenv setup for a uv project.. and Python 3.9 recommended? ), and code with separators, emojis and the like. I suspect it's not a good idea to cache the load_models call, since you're likely to want to add models while the app is running.
Please can I ask you to work through this and send something concise, written personally by you. If there are comments, make them about what you learned.
Thanks much
Ed

ed-donner:
Great - thanks so much!

PR #3340, opened 2026-06-06
Adds an accessible technical tutor for screen reader and keyboard-only Users

judithhinlung:
## Summary
This PR adds an accessible technical concept explainer for screen reader and keyboard-only users.
The tool is built twice: once using the Anthropic API for nuanced, context-aware explanations, and once using a local model, Llama 3.2, to demonstrate the same concept running entirely offline without an API key. 
The explanation format is deliberately structured for linear audio consumption: a one-sentence plain language summary first, followed by an analogy that uses no visual references, followed by a step-by-step breakdown using numbered points that a screen reader announces clearly, followed by a concrete real-world example grounded in something a blind or low vision user would encounter in daily life, and ending with one follow-up question the user might want to ask next. 
The tool avoids spatial metaphors ("as shown above," "the box on the left," "the diagram below"), color references used as the sole means of conveying information, and ASCII art or table-based layouts that become noise when read aloud.
NOTE: I decided not to use a while loop to continuously take in question inputs because the accessibility of output from a running cell is not reliable. This behavior was found while testing with VSCode and the VoiceOver screen reader.

## Testing Done
The utility successfully calls Claude and Ollama to answer a technical question.

### DIFF
--- a/week1/community-contributions/accessible-technical-concept-explainer.ipynb
+++ b/week1/community-contributions/accessible-technical-concept-explainer.ipynb
@@ -1,0 +1,346 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "id": "0942d445",
+   "metadata": {},
+   "source": [
+    "# Accessible Technical Concept Explainer\n",
+    "\n",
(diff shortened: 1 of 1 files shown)


ed-donner:
Great - thanks so much!

PR #3348, opened 2026-06-08
added improved rag solution using supabase

aswamina:


ed-donner:
Great - thanks so much!

PR #3363, opened 2026-06-12
recipe_maker

jhaak1:
I used 2 different AIs to create new recipes.  The first one pulls recipes from the internet, and the second one uses those recipes as inspiration to create a new recipe.


ed-donner:
Great - thanks so much!

PR #3366, opened 2026-06-12
Added day 4 exercise

hmnhGeek:
Added functionalities to the airline chatbot include:

1. Setting the price of a city.
2. Emptying the entire table if required.
3. Getting the prices of all the cities in one go.
4. All this is done in a pythonic way without using if-else for each method.


ed-donner:
Great - thanks so much!

PR #3367, opened 2026-06-12
Created synthetic data generator.

antonKenderov:
## I used:
- Gemma for open-source generation of the data.
- Qwen 3.5 for giving the formatted output as an object.


ed-donner:
Great - thanks so much!

PR #3378, opened 2026-06-14
Exercise of Week 3 - World Cup Result Generator

farrasezra05:
Using open-source models to create a hypothetical dataset generator, using World Cup matches as a use case.


ed-donner:
Would it be possible to save that Colab without outputs to reduce file size? Thanks so much

farrasezra05:
Updated the file to remove the outputs, thanks Ed!

ed-donner:
Ah.. this has all the data itself in there, so it's 32 files and 3.5K lines of code.. it would be more common to put the data somewhere central (like google drive) and have a README telling people how to download the data. Would that be possible? Thanks so much

PR #3389, opened 2026-06-16
Add week1 lab1 solution by PGC

pgcmerc031:


ed-donner:
Great - thanks so much!

PR #3393, opened 2026-06-18
showing plot of embeddings for better understanding

Mechantchulo:


ed-donner:
Great - thanks so much!

PR #3394, opened 2026-06-18
Adds accessible design review panel

judithhinlung:
## Summary
This change adds an accessible design review panel.
A UX designer pastes a component description or design spec. Three LLM reviewers each take a different perspective: one reviews from a screen reader user's perspective, one from a keyboard-only user's perspective, and one from a cognitive disability perspective. A fourth LLM synthesizes the three reviews into a prioritized list of design changes with WCAG criterion references. Useful for design teams who want accessibility feedback before a component is built.

Sessions can be saved to disk and analyzed over time: a built-in trend pipeline tallies the most frequently flagged WCAG criteria across all saved sessions and asks Claude to produce a plain-language training briefing — helping teams spot systemic gaps in their design practice and prioritize what to learn next.

## Testing Done
Used the sample inaccessible component design to verify that violations were flagged by the three LLM reviewers and synthesized into a priorities list by the fourth LLM.


ed-donner:
Great - thanks so much!

PR #3398, opened 2026-06-19
Adds ARIA role and attribute explainer

judithhinlung:
## Summary
This PR adds an ARIA Role and Attribute Explainer using the Gradio chat UI. The developer pastes an HTML snippet with ARIA roles or attributes into the chat, and Claude-sonnet-4-6 uses turn-by-turn dialog to guide the developer in the correct implementation.
The agent reviews the markup, identifies any invalid role and attribute combinations, missing required attributes, or redundant ARIA that duplicates native semantics, and explains each issue in plain language with a corrected version. It then asks follow-up questions about the component's dynamic behavior — does the expanded state change, does the selected state update, does focus move on activation — and generates the JavaScript needed to keep ARIA states in sync with the UI. The conversation continues until the developer has a complete, correct, and accessible implementation.

## Testing Done
Pasted one of the sample invalid implementations into the chat and received guidance, a corrected snippet, and a followup question that leads to correct implementation.


ed-donner:
Great - thanks so much!

PR #3403, opened 2026-06-20
Week 2, Day 4: Airline Ticket Pricing Assistant with LLM Tool Calling

sbm4034:
Demonstrates how to use OpenAI's function calling API to create an airline ticket pricing assistant. The assistant can retrieve and update ticket prices based on user requests, showcasing the implementation of tool routing and the strategy pattern with Gradio UI integration.


ed-donner:
Great - thanks so much!

PR #3409, opened 2026-06-21
Day1 Solution by sroy

AnneDroidd:
- added sroy dir in community_contributions directory
- added day1.ipynb file, my own solution
- added scraper.py for dependencies


ed-donner:
Great - thanks so much!

PR #3420, opened 2026-06-23
add week2 end exercise solution

Mon711:
created a Django chatbot with tool use


ed-donner:
Great - thanks so much!

PR #3421, opened 2026-06-23
Week 2 Day 3

jhaak1:
I made a career coach chatbot using GPT-OSS through Groq.


ed-donner:
Great - thanks so much!

PR #3424, opened 2026-06-24
Add Hritik day 1 solution notebook

Hritikd:
Add Hritik day 1 solution


ed-donner:
Great - thanks so much!

PR #3427, opened 2026-06-24
Add email summarization and subject line generation using Ollama

Pratha-Joshi:
LLM-powered email summarization and subject line generation using Ollama.


ed-donner:
Great - thanks so much!

PR #3428, opened 2026-06-24
Add Day 4 CLI chatbot with memory

JatinThapa:
## The goal of this mini assignment was to understand how Large Language Models (LLMs) can be accessed locally through Ollama and how conversational memory works in AI chat applications. ##


ed-donner:
Great - thanks so much!

PR #3431, opened 2026-06-25
fix: Correct description of Gradio frontend app framework

Haosik:
Not React but Svelte


ed-donner:
Very true! Silly me. Thank you.

PR #3437, opened 2026-06-26
Added brochure PDF generation with proper naming, instead of raw mark…

divyamt50:
…down code generation saving the extra effort


ed-donner:
Great - thanks so much!

PR #3439, opened 2026-06-26
Adds an accessible audio technical concept explainer

judithhinlung:
## Summary
A conversational tool that takes a technical question or concept and explains it in a way that is optimized for screen reader and keyboard-only users — people who cannot rely on diagrams, visual metaphors, or "as you can see" phrasing to understand technical ideas. 
The tool is built twice: once using the Anthropic API for nuanced, context-aware explanations, and once using a local model, Llama 3.2, to demonstrate the same concept running entirely offline without an API key. 
The explanation format is deliberately structured for linear audio consumption: a one-sentence plain language summary first, followed by an analogy that uses no visual references, followed by a step-by-step breakdown using numbered points that a screen reader announces clearly, followed by a concrete real-world example grounded in something a blind or low vision user would encounter in daily life, and ending with one follow-up question the user might want to ask next. 
The tool avoids spatial metaphors ("as shown above," "the box on the left," "the diagram below"), color references used as the sole means of conveying information, and ASCII art or table-based layouts that become noise when read aloud.

## Audio Upgrade

This version adds text-to-speech audio output as the primary delivery mechanism rather than an afterthought. The entire purpose of this tool is to make technical explanations work without visual output, so making text-to-speech the default closes the feedback loop the tool was designed for: an explanation that sounds good when spoken is fundamentally different from one that merely avoids visual metaphors on the page, and the only way to verify that difference is to hear it.

The developer asks a technical question, the tool generates an explanation structured for linear audio consumption, and then speaks it aloud using OpenAI's TTS API so the user can evaluate whether the phrasing, pacing, and structure actually work when heard rather than read.

## Testing Done
- Asked the explainer a question and tested that both Claude and Ollama responded with different voices. 
- - Verified that autoplay has been disabled to give screen reader users control over when to start and stop playback.


ed-donner:
Great - thanks so much!

PR #3445, opened 2026-06-28
for day1, I changed the codes for local ollama use and the exercise a…

imphe1218:
The changes are the example codes I have replaced them with Ollama instead of connecting to a frontier model with an API Key. The exercise before the Scrapping I did but not the Scrapping.


ed-donner:
Great - thanks so much!

PR #3446, opened 2026-06-28
day 2, i have modified day 1 file on the web summarization codes to u…

imphe1218:
day 2, i have modified day 1 file on the web summarization codes to use ollama but still didnt do the web scrapping thing.


ed-donner:
Great - thanks so much!

PR #3448, opened 2026-06-28
Daniel Hernandez Week3 exercise

dhernandezgt:
Please check exercise for week3.


ed-donner:
The poetry sentiment colab looks great to me

PR #3452, opened 2026-06-29
Created a tutor using Ollama

Pratha-Joshi:
Built a tool using Ollama that takes any question and responds with an explanation.


ed-donner:
Great - thanks so much!

PR #3468, opened 2026-07-01
Kanika ai tutor

kanikasecl7:
An AI Tutor using Ollama, brochure practice and 3 chatbots interaction with some side test fiels


ed-donner:
Great - thanks so much!

PR #3475, opened 2026-07-02
Add Week 3 community contribution

Adityaz23:
### This is the week-3 assignment which we have completed using openrouter and huggingface transformers.
**Also created the gradio ui for it aswell.**


ed-donner:
Great - thanks so much!

PR #3487, opened 2026-07-03
Day 1: UV install update

amitg0161:
Not showing the vir env for VS code, so followed the instruction from the link and removed the uv lock and installed it.

Raising PR. Please review if found useful


ed-donner:
Hmm I'm not following you.. you want me to merge this and change the uv.lock for the course?

PR #3502, opened 2026-07-07
Add Week 1 solution with cleared outputs

oNk2r:
contains week1-exercise-solution 
thanks!


ed-donner:
Great - thanks so much!

PR #3508, opened 2026-07-08
Update API key validation in day2 notebook

asis-das:
Updated the API key validation check in week1/day2.ipynb.

Changes:
- Updated the expected API key prefix
- Updated the validation message


ed-donner:
You'd need to look for either key prefix - either are valid - and actually it looks like you didn't change the print message? thanks

PR #3509, opened 2026-07-08
Added lab file and week1 exercise with Anthropic KEY

vlalez:
Added day5.ipynb file and week1 exercise with Anthropic Key and Anthropic syntax


ed-donner:
Great - thanks so much!

PR #3522, opened 2026-07-10
Optimus Prime Website Summarizer

CodeMaestroMS:
Implements a website summarization tool powered by a local Llama model via Ollama, with Optimus Prime as the persona.

- Scrapes website content and summarizes it in the voice of Optimus Prime
- Uses Ollama (`llama3.2`) as the local LLM backend
- User and system prompts themed around Transformers/Cybertron


ed-donner:
Great - thanks so much!

PR #3524, opened 2026-07-10
Week 5 Day 3: A RAG pipeline for assisting with .epub books

hmnhGeek:
Place your books in knowledge-base folder and run the ingest pipeline. Then start chatting with the assistant on those books.


ed-donner:
Great - thanks so much!

PR #3526, opened 2026-07-10
Add unit test generator tool with support for multiple models

OluwadaraDaily:
# What does this PR do?

I worked on a tool that would generate unit tests from both open & closed source models.
In addition, I added a tab section to compare different output from the different models.

In this case I compared: GPT 5, Claude Sonnet 4.5, Cogito (14B), Qwen 3.5 (9B), DeepSeek R1 (1.5B), Llama 3.2


ed-donner:
Great - thanks so much!

PR #3532, opened 2026-07-11
feat: day1 week1 of work.

rakshit1806:
Used Ai news as website scraper endpoint and also use free llama model3.2 from llama


rakshit1806:
First commit dealt with switching to llama model.
Second commit was the exercise with custom **system prompt and user prompt**., that was left out in first commit merge

ed-donner:
Great - thanks so much!

PR #3533, opened 2026-07-11
Week1 day1 and day2 to community-contributions

auroraa-lola:


ed-donner:
Great - thanks so much!

PR #3534, opened 2026-07-11
Week 1 Exercise: Feature Creep Simulator

frankie-hung:
## What is this?

A small game I built called **"Feature Creep Simulator"** — each round, the LLM 
adds a new element that must logically connect back to every previous element in 
a growing list, and self-rates how coherent that connection is. The idea is to see 
how gracefully (or not!) a model handles increasing complexity, and to watch 
"coherence" decay round by round.

## What I learned / built

- Practiced the core stateless API pattern — re-sending the full `messages` 
  history each round to simulate memory
- Wrote a system prompt with a concrete scoring rubric to reduce the model's 
  self-grading bias
- Added simple Python-side safeguards so the game loop doesn't depend on the 
  LLM to police its own rules:
  - Ends automatically if the Coherence Score drops too low
  - Ends if the user hits Ctrl+C or types "stop"
  - Ends if the LLM gives a malformed/invalid response

## Fun bugs I ran into along the way

- The model sometimes declared "Force End" *because* the score was high, 
  not low 😅
- With `gpt-4.1-nano`, it occasionally just refused to continue with no reason
- It would sometimes forget to reprint the running list — tightened the system 
  prompt wording to fix this

## Files added

- `week1-exercise/feature_creep_simulator.ipynb`

Excited to be contributing to this repo — thanks for putting together such a 
great course! 🙌


ed-donner:
Great - thanks so much!

PR #3551, opened 2026-07-14
Add abstractserenity/Jess's day 1 notebook contribution.

abstractserenity:
Adding only the notebook with my day 1 contributions including a funny addition and use of the LLM taking on the persona of a typical Aussie man. I also used my Claude api instead of OpenAI so those changes are also present.


ed-donner:
I believe merging this PR would delete week1 day1 from the repo because you've renamed it?

PR #3555, opened 2026-07-15
A small project to get clothes advice

sergey-misuk-personal:
A small project to get clothes advice for a week in a specific city using open-meteo.com API and local Llama model


ed-donner:
Great - thanks so much!

PR #3557, opened 2026-07-15
By atul-cs Mini-project multibot group chat

atul-cs:
Made a mini-project on multi-AI group chat, with user Interaction.
Users Can interact.
Can run as many LLM's or models local or on cloud.


ed-donner:
Great - thanks so much!

PR #3558, opened 2026-07-15
Add global market news summarizer

lutfikhamami:
Scrapes global market news headlines from Yahoo Finance, extracts article content, and uses an LLM via OpenRouter to generate a concise daily market briefing (sentiment, sectors in motion, key takeaways).


ed-donner:
Great - thanks so much!

PR #3567, opened 2026-07-17
Add dynamic web + PDF-aware scraper with adaptive Selenium and chunked LLM summarization

tchatty:
### A smart ingestion pipeline for modern web content (HTML + JS + PDF) with scalable LLM summarization.

This PR adds a project that extracts and summarizes content from modern websites using a multi-stage, adaptive pipeline.

It supports:
- Static HTML scraping using `requests`
- JavaScript-heavy pages via Selenium fallback
- Detection and extraction of embedded or linked PDF content
- Chunk-based summarization for long documents


ed-donner:
Great - thanks so much!

PR #3574, opened 2026-07-18
Story teller

sebahadin:
This pull request introduces an multi-modal storytelling application built with Gradio and the OpenAI API. The app uses tool calling to generate a story outline, create AI-generated images for each scene, and narrate the story.


ed-donner:
Great - thanks so much!

PR #3577, opened 2026-07-19
Added a tool that generates comments.

Pratha-Joshi:


ed-donner:
Great - thanks so much!

PR #3579, opened 2026-07-19
Anurag rawat/week 1 day 2

Anurag-NURA:
**Ollama-based grammar assistant (`day_2.ipynb`):**
- Provides instructions to install and verify Ollama and download the Llama 3.2 model.
- Installs the required Python package, sets up the model, and constructs a similar prompt for grammar correction.
- Sends the prompt to the local model and prints the response with corrections and explanations.


ed-donner:
Great - thanks so much!

PR #3584, opened 2026-07-21
Three way conversation example

rishi1011:
Adds a multi-agent conversation example where three football fans (Barcelona, Real Madrid, and Atlético Madrid) debate using role-specific prompts and shared conversation history.


ed-donner:
Great - thanks so much!

PR #3587, opened 2026-07-21
Add Afeez week1 lab 1&2 solution. I used Deepseek & llama3.2

akpopoola:
Suggestion to add this book recommender to the community_contributions for week1


ed-donner:
Great - thanks so much!

PR #3599, opened 2026-07-23
Added Week 1-Day 1 Community Contribution - Scam Detector 

tobilobasalawu:
A simple Python app that uses OpenAI to check if a message might be a scam.


ed-donner:
Great - thanks so much!

PR #3600, opened 2026-07-23
Removed stray NTFS Zone Identifier, and added a gitignore entry 

steve-potts:
`week1/community-contributions/Anurag_Rawat/week1/day_1.ipynb:Zone.Identifier`
contains a colon, which is not a legal character in Windows filenames. Cloning
on Windows fails partway through checkout:

    error: invalid path 'week1/community-contributions/Anurag_Rawat/week1/day_1.ipynb:Zone.Identifier'
    fatal: unable to checkout working tree

The clone "succeeds" but the working tree is empty and every tracked file shows
as deleted in `git status`, which is confusing if you don't know the cause.

The file itself is an NTFS alternate data stream (Windows' mark-of-the-web
marker) that got flattened into a regular file, most likely when the notebook
was copied through WSL. It carries no useful content and nothing references it.

Note: this can't be deleted through GitHub's web UI — the delete endpoint
returns a 422 on that path. It needs removing from a clone on macOS or Linux,
or via `git update-index --force-remove` on Windows.


steve-potts:
Ed Fixed this already

PR #3606, opened 2026-07-25
using local llm SmolLM2-135M-Instruct-Q4_K_M.gguf running on AntiX Li…

imphe1218:
…nux, a debian 13 code base flavor with very minimal and non NVIDIA GPU


ed-donner:
Great - thanks so much!

PR #3607, opened 2026-07-25
Week3 - Jitesh - Solution using gradio for data generator

jjmanglani01:
Require python 3.1
OPANAI_API_KEY
ollama for local inference


ed-donner:
Great - thanks so much!

PR #3619, opened 2026-07-26
Week 2 excercise updated

Pavankumarmanagoli:
Added week 2 excercise and I have not changed anything in .gitignore its same.


ed-donner:
please remove that file from the PR.. even though it appears to be deleting a line and adding the same line back in, it's important to have a sensible, prudent approach to change management. There are a large number of people enrolled on this course. If this change results in a merge conflict with 1% of them when they do a git pull, count how many emails I'll get..

PR #3620, opened 2026-07-26
Community pr week1: Ollama API Key

dapulache:
Hello Ed,

Here is my community contribution to week1. In it, I was able to do it all for free using Ollama but I took it one step further. Instead of using the Ollama app and hosting an API endpoint virtually. I generated an Ollama API key, added it to the .env file (don't worry it is not in the PR) and changed the code around a bit so that it uses the Ollama API Key to call models. This is basically the same thing as paying ChatGPT to let you generate a key but this solution is free. 

Thank you for your course. I am really enjoying it!

All the best!


ed-donner:
Yes this looks like you're using Ollama Cloud, which is similar to an OpenRouter.. thank you

PR #3623, opened 2026-07-27
Add Tigeress Week1 Exercise and  self practice zillow Projects

Tigeressgithub:
Add Tigeress Week1 Exercise and  self practice zillow Projects which will be improved after learning more about LLM


Tigeressgithub:
fixed the conflict

ed-donner:
Great - thanks so much!

PR #3631, opened 2026-07-29
Add LLM Cost Meter: token and spend tracking across course providers

Nishant-444:
The course keeps total API spend to a few dollars but never shows what an individual call costs. This adds a small meter that reports tokens and estimated spend.

The course drives OpenAI, Anthropic, Gemini, DeepSeek, Groq, Grok, OpenRouter and Ollama through the same OpenAI client with a different base_url, so wrapping chat.completions.create instruments all of them at once. wrap() patches the client in place and returns it, so existing code is unchanged.

Reasoning tokens are broken out separately, which makes the cost of reasoning_effort in week2/day1 visible for the first time.

Token counts come from the provider and are exact. Costs are estimates from a dated table that can be overridden; cache and batch discounts are not modelled, so a run never costs more than reported. Local models count as free and unknown models are reported as unpriced rather than silently zero.

No new dependencies.

### DIFF
--- a/community-contributions/Nishant-444/llm_cost_meter/README.md
+++ b/community-contributions/Nishant-444/llm_cost_meter/README.md
@@ -1,0 +1,63 @@
+# LLM Cost Meter
+
+Tracks token usage and estimated API spend across the providers used on this course.
+
+The course keeps total costs to a few dollars, but never shows what an individual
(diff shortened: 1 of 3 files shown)


ed-donner:
Great - thanks so much!

PR #3632, opened 2026-07-29
Fix Course Navigator crash on Python 3.11

Nishant-444:
app.py used a backslash inside an f-string expression, which is only valid from Python 3.12 (PEP 701). pyproject.toml declares requires-python = >=3.11, so students on 3.11 hit a SyntaxError at import and the app never starts.

Build the blockquote string before interpolating it instead.

### DIFF
--- a/community-contributions/Nishant-444/course_navigator/app.py
+++ b/community-contributions/Nishant-444/course_navigator/app.py
@@ -142,7 +142,10 @@ def explore_search(query, top_k):
             if file_type == "python" or (file_type == "notebook" and meta.get("cell_type") == "code"):
                 md_output += f"```python\n{content}\n```\n\n"
             else:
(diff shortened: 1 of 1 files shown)


ed-donner:
But the version of python is pinned to 3.12.12 in .python-version

Nishant-444:
You're right — with .python-version pinned to 3.12.12, anyone following the standard uv setup will never hit this, so "crash" overstates it in my original description.
The reason I noticed it: pyproject.toml declares requires-python = ">=3.11", and on 3.11 that line is a hard SyntaxError at import (PEP 701 only landed in 3.12). So it's really a small inconsistency between the declared floor and the pinned version, not a bug students will run into.
Happy either way — merge it as a tidy-up, or close it if you'd rather leave the file alone. Not blocking on my end.

PR #3640, opened 2026-07-30
Add Abdullah's week1 day1 solution using Ollama

Abdullah-Program:
Website summarizer using Ollama (llama3.2) instead of OpenAI, following Week1 Day1 lab.


ed-donner:
Great - thanks so much!

PR #3643, opened 2026-07-30
Stefan week 4 day 4 replace leibniz by chudnovsky

GordonFreeman-17:
As suggested the PR to replace leibniz by chudnovsky when running the python to c++ translation tool as community contribution. I added the gauss legendre code as bonus :-).


ed-donner:
Such a joy! Thank you Stefan.. brilliant

PR #3645, opened 2026-07-30
A hybrid RAG pipeline utilizing keyword search and vector search

indrajitsg:
A retrieval-augmented generation pipeline over the Insurellm knowledge base, using **hybrid retrieval**, **query rewriting**, and **LLM reranking**.

This module is built to utilize the Ollama model **ornith:9b**. With minor modifications
an OpenAI or Anthropic model can also be used.


ed-donner:
Great - nicely solved! Thanks for the contribution

PR #3659, opened 2026-07-31
A tool that answers user questions with 3 different models

akpopoola:
implemented the AI assistant to answer programming and LLM-related questions. I set it up to use 3 LLMs (deepseek, llama3.2, and phi3) to compare the answers.


ed-donner:
Great - thanks so much!

PR #3670, opened 2026-08-02
added a bookstore assistant equiped with tool calling capabilities

s-ig-ma:
This is a simple bookstore assistant project that is equipped with tool calling capabilites. It works as follows. 
-User requests to buy, rent or sell, a book.
-The appropriate tool is called which makes an api call to booksrun api to get the relevant prices for that book, using the isbn number.


ed-donner:
Great - thanks so much!

PR #3678, opened 2026-08-04
Add Model Arena: GPT vs Claude voting notebook

CodeMaestroMS:
A simple Gradio app that asks GPT and Claude the same question, streams both answers side by side, and lets you vote on which one answered better. Built on the Week 2 Day 2 Gradio lesson.


ed-donner:
Great - thanks so much!

PR #3707, opened 2026-08-10
Add week1 day1 community contribution

mostafa-hatem1:
created a helpful career assistant that reviews your CV against a job description, and gives you a rating on how much of a fit you possibly are.


ed-donner:
Great - thanks so much!

PR #3709, opened 2026-08-11
feat: Add 3-way engineer conversation notebook

scryptwiz:


ed-donner:
Great - thanks so much!

PR #3711, opened 2026-08-11
Week 1 exercise: technical question explainer (GPT + Ollama)

Alex-Huaracha:
End of week 1 exercise: a tool that answers technical questions with streaming, using both gpt-4o-mini and llama3.2 via Ollama.

The notebook keeps a small model registry that pairs each model with its client and its own system prompt. llama3.2 gets a shorter prompt with a one-shot example, because the longer instruction-heavy prompt written for gpt-4o-mini makes it drift.

Only adds one file under community-contributions. Outputs are cleared.


ed-donner:
Great - thanks so much!

PR #3714, opened 2026-08-12
Add Week 3 exercise: Synthetic Q&A Dataset Generator

wachira7:
Builds on the Week 1/2 technical tutor by generating a synthetic Q&A training dataset for it, using three models in three distinct roles so "diverse outputs from a variety of models" is structural rather than cosmetic:

- One model writes the questions for a given topic/difficulty
- A different model writes the answers, so it isn't grading its own homework
- A third model judges each pair (1-5 score + notes) and separately assesses what difficulty level the pair actually reads at, flagging mismatches against what was requested

Only pairs at or above a user-set minimum score are kept in the final JSONL export; the full generation table (including filtered-out and mismatched rows) stays visible in the UI for inspection.

**Gradio UI:** topic/difficulty/count inputs, independent model pickers for each of the three roles, live per-row progress during generation, a tuned-width dataframe preview, and a JSONL download of the kept pairs.

Smoke-tested the generation pipeline end-to-end (question generation, answer generation, judging, JSON parsing, difficulty-match detection, file export) and manually verified the Gradio UI.

### DIFF
--- a/week3/community-contributions/synthetic_qa_dataset_generator_waruts.ipynb
+++ b/week3/community-contributions/synthetic_qa_dataset_generator_waruts.ipynb
@@ -1,0 +1,359 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "id": "341f0ce3",
+   "metadata": {},
+   "source": [
+    "# Week 3 Exercise: Synthetic Q&A Dataset Generator\n",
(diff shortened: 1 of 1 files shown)


wachira7:
FYI Ed

ed-donner:
Great - thanks so much!

PR #3735, opened 2026-08-16
Week2 day1 exs last

bezaleluriel:
Added 3 versions


ed-donner:
Great - thanks so much!

PR #3736, opened 2026-08-16
feat(week1): add GPT→Llama lab sidekick exercise

varunsmhatre:
Pipeline: GPT explains the concept, Ollama Llama writes a matching example.


ed-donner:
Great - thanks so much!

PR #3755, opened 2026-08-20
Add Week 2 AI Tutor conversation context

damaniayesh:
## Description

This contribution upgrades the Week 1 AI Tutor by adding conversation context.

- Added the previous two conversations as background information
- Added a new technical question
- Combined the conversation history into one user prompt
- Reused GPT-5-nano and Llama 3.2 with the same context
- Added streaming for both model responses


ed-donner:
Great - thanks so much!

PR #3758, opened 2026-08-21
Week 2: helpdesk assistant with specialist routing and ticketing

Filiup:
Week 2 community contribution: a hosting-provider helpdesk assistant that routes every technical question to one of three specialist LLMs instead of answering on its own.

**What it does**

- Customer-facing assistant has no technical knowledge — it must call `ask_specialist` (network, billing, security) for every technical answer.
- Each specialist keeps its own conversation history, so follow-ups only carry the new fact instead of restating the case.
- Specialists answer in a fixed ASSESSMENT / VERDICT / NEXT STEP format; `NEEDS_HUMAN` makes the assistant create a ticket.
- Tickets are stored in SQLite (`create_ticket`, `get_ticket`), so a customer can come back with a ticket ID and the assistant picks up the context.
- Gradio UI shows the customer chat plus all three specialist threads side by side, and the reply is spoken via `gpt-4o-mini-tts`.

Files: `week2/community-contributions/helpdesk-assistant/week2_helpdesk_assistant.ipynb`


ed-donner:
Great - thanks so much!

PR #3768, opened 2026-08-23
Added Jokes AI -Scrapper

Shraghav:
A simple jokes ai scrapper to test my understanding on scrapping the web and generating jokes based on that


ed-donner:
Great - thanks so much!

PR #3771, opened 2026-08-23
Add Sajjad week 1 contribution

jonayeds:
Added Gen Ai mini project (Portfolio Reviewer) using GEMINI api. 
Project Description: Portfolio reviewer assistant. It fetches all relative routes from the portfolio website and combine the contents from the website in a structured way. Then Gemini model reviews the contents and projects in detail. Assistant makes a judgement on the current career trajectory and suggest detailed review to grow in career. And The assistant also looks into the content and suggest improvements on the written contents that would be more relevant to the skillset of the client.


ed-donner:
Great - thanks so much!

PR #3772, opened 2026-08-23
Add week2 community contributions: course tutor chatbot and brochure generator/translator UI

mostafa-hatem1:
made a socratic tutor for this course specifically, that thinks along side me instead of giving the answer straight away. Also added on top of the brochure generator from the previous week to include a translator feature with Gradio serving up the UI


ed-donner:
Great - thanks so much!

PR #3774, opened 2026-08-23
Week 5 Performance Improvement

clinton2111:
Added hierarchical RAG technique to summarise data and also increased the number of generated questions from a single rewrite to 5 ( original + 4 new)


ed-donner:
Great - thanks so much!

PR #3775, opened 2026-08-23
add challenge notebook

NaKamize:


ed-donner:
Ahh there's a way to export a notebook from colab without it including the 'widgets' and other outputs so that it's less than 3,700 lines.. if you're able to do that it would be v much appreciated.. thanks so much

NaKamize:
> Ahh there's a way to export a notebook from colab without it including the 'widgets' and other outputs so that it's less than 3,700 lines.. if you're able to do that it would be v much appreciated.. thanks so much

Oh, my bad. I have pushed fixed version which is hopefully fine.

ed-donner:
Great - thanks so much!

PR #3781, opened 2026-08-25
Add 3-way LLM chat implementation with logical conclusion of chat

Javedfaruquee:
Add 3-way LLM chat implementation, using GPT, Anthropic and Gemini and also ensuring the chat has logical conclusion.


ed-donner:
Great - thanks so much!

PR #3784, opened 2026-08-27
Add cleaned wk3 and wk4 contributions

auroraa-lola:


ed-donner:
Great - thanks so much!

PR #3800, opened 2026-08-30
day 05 and week 1 solution

jaffasunny:


ed-donner:
Great - thanks so much!

PR #3802, opened 2026-08-30
Add: week 2 day 4 exercise

rishi1011:
Tasks
 - Added set price function to the airline assistant.
 - Called tool methods in a pythonic way by passing in the arguments to its relevant methods without checking the method name.


ed-donner:
Great - thanks so much!

PR #3803, opened 2026-08-30
This is our approach with some new ollama models, cloud models and cl…

Adityaz23:
…osed source models


ed-donner:
Great - thanks so much!

PR #3804, opened 2026-08-30
Add Ranuga's lab 2 solution (cleared outputs)

ranuga-anthony:
Commit summarization homework using ollama


ed-donner:
Great - thanks so much!

PR #3808, opened 2026-08-31
Hi Mr Donner, here's my contribution for the week 1 day 1 challenge

NedytheCoder:
I modified the system prompt to be a bad boy who gets girls effortless, and had it give me a line to use on a newly added girl on facebook lol


ed-donner:
Gosh! I realize you're kidding around.. but this seems like the kinda thing that might be controversial, particularly given the number of people from all over the world on this course! If someone takes this out of context that doesn't know you, they might react negatively. I don't suppose I could ask you for a slightly less controversial example...? But thank you for taking the time to contribute..

NedytheCoder:
Good day Mr. Donner. You are right. I need to be more careful with
controversial themed contributions. I apologized. I've updated the code to
be an Electrical Engineer as I have a background in that field. Thank you!

ed-donner:
Thank you! Please clear outputs for me.. thanks

PR #3812, opened 2026-08-31
Added simple tooling usecase (word counter, character occurence) and simple 3 way chat orchestrator

Shraghav:


ed-donner:
Great - thanks so much!

PR #3813, opened 2026-08-31
Add Transformer Playground

Pragati241299:
# 🤗 Transformer Playground

**Transformer Playground** is a hands-on project exploring how pretrained Transformer models
actually work, from high-level pipelines down to raw next-token prediction — built with
Hugging Face Transformers and PyTorch, running on a Colab GPU.

Rather than only using high-level APIs, this notebook gradually goes deeper into the Transformer
workflow: what actually happens between typing text and getting a prediction back.

### 📌 What's covered

1. **Pipelines** — sentiment analysis, NER, zero-shot classification, and text generation, all through Hugging Face's one-line API
2. **Tokenizers** — comparing how BERT and GPT-2 split identical text completely differently
3. **Models** — loading a model directly with `AutoModel`, inspecting raw output instead of a pipeline's finished answer
4. **GPU inference** — measuring real GPU vs. CPU speed on the same model, same input
5. **Next-token prediction** — watching a model weigh several candidate next words with real probabilities, one token at a time

### 🛠️ Tech stack
- Python
- Hugging Face Transformers
- PyTorch
- Gradio (interactive demo)
- Google Colab (T4 GPU)

### DIFF
--- a/week3/community-contributions/pragati_transformer_playground/transformer_playground.ipynb
+++ b/week3/community-contributions/pragati_transformer_playground/transformer_playground.ipynb
@@ -1,0 +1,55 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "id": "a41da8fb",
+   "metadata": {},
+   "source": [
+    "# 🤗 Transformer Playground\n",
+    "\n",
+    "\n",
+    "### Project Description\n",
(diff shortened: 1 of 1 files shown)


ed-donner:
Great - thanks so much!

PR #3821, opened 2026-09-02
Week1 contribution

vmadhoobiz:
Created a lab that reads from a free open API that provides the current location of the International Space Station. The code retrieves the json response extracts out the latitude and longitude. It then creates a prompt to use the coordinates and behave as a travel agent to create a sales pitch to visit that location.


ed-donner:
Great - thanks so much!

PR #3826, opened 2026-09-03
Week-2: Tried a grounded response with adding skills harness instruction prompt for presenting the product

Gokul-kishan:
Contribution for e-commerce business sample with bunch of products for grounded response and skills to be used to presentation for LLM response


ed-donner:
Great - thanks so much!

PR #3828, opened 2026-09-03
Changed the system prompt a bit and added the functionality to set prices for city

SANDEEPNEGI07:
I have changed the system prompt a bit, so that the assistant is only focused to what this flight business offers and ignore user prompt if it's about something other than flights details.


ed-donner:
Great - thanks so much!

PR #3829, opened 2026-09-03
Add week4 automatic test generator community contribution

mostafa-hatem1:
created a notebook that automatically creates tests for your sample python function or user functions, and executes them and produces a report for it. it also does one bug fixxing loop in case of failed tests. gemini-3.5-flash-lite and gpt-oss-120B (GROQ) are options to be used.


ed-donner:
Great - thanks so much!

PR #3831, opened 2026-09-04
add UI-driven synthetic data generation from JSON schema

hk22sync:
Added schema-driven synthetic data generation using Llama 3.2 3B Instruct.

- Allow users to define JSON schemas through the UI
- Generate realistic fictional records based on schema fields
- Use field keys, types, and descriptions to guide data generation
- Extract and parse JSON from model output
- Support generation of 20 synthetic records per request


ed-donner:
Great - thanks so much!

PR #3833, opened 2026-09-04
Add ragforge

Pragati241299:
A RAG assistant that answers questions about frontend engineering docs and team process documents. Built two ways — with LangChain and without it — so they can be compared side by side, plus real evaluation metrics to measure which one actually works better.


ed-donner:
Great - thanks so much!

PR #3836, opened 2026-09-06
Week 8: verify the price of the selected deal with a web search

kirstenbeekhuis:
Deals in week 8 are ranked on the estimate minus the asking price, and that estimate comes from the
ensemble: a frontier model doing RAG over Amazon Reviews 2023, the fine-tuned pricer from week 7, and
the neural network from week 6. All three sit in the same 2023 price world, and the test set is a
slice of the same shuffled data, so those numbers say how well the models predict 2023 prices. The
prompt carries no date, so the models cannot answer for another year. The ranking between products
carries over to today, the price level does not, which means a deal can clear the threshold on the
gap between two years rather than on a real saving.

This adds one step to the planner, in week8/community_contributions/price_verification/. Once the
best deal of a run has been picked, it looks up what that product currently sells for elsewhere with
the OpenAI Responses API and the web_search tool, and measures the discount against that price. The
threshold test then runs on the looked up price. Five deals are estimated and only the winner is
looked up, so it is one search per run. Looking up every deal instead is not an alternative: a lookup
needs one identifiable product and costs a search per deal, while the ensemble can price everything,
so it stays as the thing that ranks the candidates.

It only works where one specific product can be identified at another retailer: the same model
number, the same condition, sold by the retailer itself, and a price paid today rather than an MSRP
or a promotion. Those conditions are set in the system prompt and repeated in the field descriptions
of the schema, and every listing carries a URL. When the product cannot be pinned down that way the
lookup returns found=false, the estimate stays and the deal is handled as before. It measures against
the lowest valid listing rather than an average, because the question is whether the same product can
be bought cheaper somewhere else.

Measured on 2026-09-06, one run each. A mini PC advertised at $329 was estimated at $510.29, a
discount of $181.29 and over the threshold of $50; the lowest listing online was $363.53, which makes
the real saving $34.53 and drops it below the threshold. A signage display from a brand with no
second retailer returned found=false, so the estimate stayed and the deal was handled as before.

The lookup is a draw in itself: the same soundbar returned $599.99, $909.99 and $699.99 over three
lookups at three retailers. Within one lookup the listings are consistent, between lookups they are
not. How often a lookup succeeds over a longer series has not been measured.

Three files, no notebooks, nothing outside community_contributions. VerifyingPlanningAgent subclasses
PlanningAgent, so nothing in week8 changes.

### DIFF
--- a/week8/community_contributions/price_verification/README.md
+++ b/week8/community_contributions/price_verification/README.md
@@ -1,0 +1,104 @@
+# Price verification after the selection
+
+## Why
+
+The estimate comes from the ensemble: a frontier model doing RAG over Amazon Reviews 2023, the
+fine-tuned pricer from week 7, and the neural network from week 6. All three sit in the same 2023
+price world, and train, validation and test are slices of the same shuffled data, so the week 6 and
+week 7 numbers say how well they predict 2023 prices, not today's. The prompt carries no date, so
+nothing in the chain can answer for another year, and more training does not fix it, because the
+information is not in the data. What carries over to today is the ranking between products, not the
+price level.
+
+That matters for week 8, because a discount is the estimate minus the asking price. If the estimate
+sits at a 2023 price level, so does the discount, and a deal can clear the threshold on the gap
+between two years rather than on a real saving.
+
+So for products that can be identified one to one at another retailer, this checks the estimate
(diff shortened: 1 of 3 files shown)


ed-donner:
Great - thanks so much!

PR #3847, opened 2026-09-08
Tool calling project from week 2

NedytheCoder:
Hi there Engineer Donner, I just completed an LLM financial assistant that uses defined functions as tool.


ed-donner:
Great - thanks so much!

PR #3850, opened 2026-09-10
Add day 3 Python tutor chatbot

lcoreman:
For the week 2, day 3 exercise, I have created a chatbot with a Gradio interface that acts as a Python tutor using GPT-5.6 Luna. Its name is Monty. Thanks!


ed-donner:
Great - thanks so much!

PR #3864, opened 2026-09-12
Used Car Price Intelligence & Prediction System

Pragati241299:
Predicts a used car's resale price from its details - brand, year, mileage, fuel type, and
more - using traditional ML (Random Forest, XGBoost) as the actual predictor, with an LLM
used for enrichment and comparison, not as the core engine.


ed-donner:
Great - thanks so much!

PR #3865, opened 2026-09-12
Add adversarial-resistant summarizer (week1 day1 extension)

themrbads:
## Summary

Took the day1 web summarizer a step further: what happens if the page you're summarizing contains text aimed at the model instead of at a human reader?

- Baseline (no defenses) vs. a hardened version (delimited untrusted content + a second self-audit call) tested against a few local fixture pages with different prompt-injection styles.
- The actual results surprised me going in — the plain baseline resisted the two more obvious injection attempts and only got hijacked by the subtlest one (a fake "system note" asking for something plausible rather than something broken).
- The hardened version's self-audit also has a real limitation I found while testing it (flags "correctly resisted and reported" the same as "actually compromised") — written up honestly in `results.md` along with what I'd fix next.

## Test plan

- [x] Ran the notebook end to end against `openai/gpt-4o-mini` and confirmed the outputs match what's documented in `results.md`
- [x] Notebook outputs cleared before commit
- [x] Only touches `community-contributions`


ed-donner:
Great - thanks so much!

PR #3866, opened 2026-09-12
auroraa week5 exercises

auroraa-lola:
For week5 community contribution, I included the knowledge base. If not ok, I can resubmit.
Thank you for your patience.


ed-donner:
Ah yes - this is great, but if possible please put the knowledgebase somewhere central (like Google Drive) with README instructions to download it, rather than having it in the repo.. thanks so much

auroraa-lola:
Will do, going forward. Many Thanks!

PR #3872, opened 2026-09-13
completed week2 exercise

abrarwahidd:
This is my exercise in week 2


ed-donner:
It seems ike Copilot has a point there - this looks like you end up making the LLM call twice? was that intentional?

PR #3873, opened 2026-09-13
Fuji Film Recipe Recommender

jairofranchi:
Given a description of the current scene (weather, light, mood, camera/sensor), recommends a matching Fujifilm film simulation recipe scraped live from fujixweekly.com.

How it works:
1. scraper.py fetches links and page content from Fuji X Weekly's recipe hub pages (one per sensor generation — X-Trans I–V, Bayer, EXR CMOS, GFX).
2. An LLM call picks the 1–3 most relevant recipe links for the user's description (or picks one at random if "random" is mentioned), returning strict JSON and never inventing URLs.
3. The content of those candidate pages is fetched and passed to a second LLM call, which recommends the best-fitting recipe, including camera/sensor compatibility and settings, always citing the source URL, and saying so if nothing fits.

Runs locally against Ollama (llama3.2 by default via the OpenAI-compatible endpoint), so no API key is required.


ed-donner:
Great - thanks so much!

PR #3874, opened 2026-09-13
Add Ranuga's Multimodal AI assistant

ranuga-anthony:
Commit my solution to week 2 exercise: VacationAI, a multimodal AI assistant (using open source models for image generation and text to speech) that helps users plan their vacations by checking the weather in their preferred destination, searching for tourist travel attractions, and in addition setting ticket prices to destinations, and getting the ticket prices to destinations.


ed-donner:
Great - thanks so much!

PR #3876, opened 2026-09-14
chore: week-4 exercise add notebook for commenting code using an LLM model

loxator:


ed-donner:
Great - thanks so much!

PR #3878, opened 2026-09-15
Add Gradio UI for Python code commenting using LLM

hk22sync:
Added an LLM utility to automatically add meaningful comments to source code while preserving the original functionality. The LLM is instructed to return only the updated code.


ed-donner:
Great - thanks so much!

PR #3897, opened 2026-09-18
Week 2: 3-Way-Chat

s-heim-dev:
Based on the contents of week 2, I built a 3-way-chat allowing to chat simultaneously with an angel and a devil. It includes a CLI-Version which also allows to simulate a user or actively participate in the conversation, and there is a Gradio-Version offering a chatbot interface with some extra quirks like images of the chatbots and the option to change the used model during the conversation.


ed-donner:
Great - thanks so much!

PR #3899, opened 2026-09-19
Add Gemini version of Day 1 notebook

loganlee23011-vn:
Hi, today is my first day learning this course.
I am new to GPT APIs, so I tried using Gemini because it is free.
I changed the Day 1 notebook to use Gemini and tested it successfully.


ed-donner:
Great - thanks so much!

